PR #6, opened 2020-06-29
add list org servlet & connect front page to it

andrewm898:
This adds a servlet that gets a list of org entry names from the database & populates them into the 'help near me' section.

NOTES:
-For the "My Organizations" page to work with this servlet (as in it only returns the orgs that a user is the moderator of), it simply needs to be called with the query param "displayUserOrgs=true", and from a point where the user is logged in

-There are a few TODO's in here: 
      1. change to return only the approved orgs as a default instead of just all of them (left this as todo bc we havent implemented org-approval yet)
      2. Implement pagination for all displays of this data (I left this as a todo because it requires some changes in the javascript infrastructure, better to worry about that after merging this first version)


src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -24,4 +40,55 @@
...
+
+    for (Entity entity : results) {
+      requestedOrganizationNames.add((String) entity.getProperty("orgName"));
+    }
+  
+    Gson gson = new Gson();
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(requestedOrganizationNames));
+  }
 }
```
Sarah-Addo: new line

src/main/webapp/index_script.js
```diff
@@ -13,37 +13,29 @@
...
+ * This async function gets a default list of organization names to display when the dom content
+ * loads by not passing any parameters to /list-organizations
+ */
+async function renderOrganizations(organizationArea) {
+  const response = await fetch(`/list-organizations`);
+  const organizationNames = await response.json();
+  for (let i = 0; i < organizationNames.length; i++) {
+    organizationArea.appendChild(createOrganization(organizationNames[i]));
+  }
+}
```
Sarah-Addo: you could make the onload function async and just do the fetch and await in there as well as using a for each loop so its shorter

src/main/webapp/index_script.js
```diff
@@ -13,37 +13,29 @@
...
+
 function createOrganization(organization) {
   const organizationElement = document.createElement("div");
   organizationElement.classList.add("organization");
 
   const organizationNameElement = document.createElement('div');
   organizationNameElement.classList.add("organization-name");
-  organizationNameElement.textContent = organization.name;
+  organizationNameElement.textContent = organization;
```
Sarah-Addo: it would have to be organization.name because organization is an object

Sarah-Addo (changes requested):

src/main/webapp/index_script.js
andrewm898: My thoughts in breaking this off to be its own function are that in the future we can just add a single parameter to it that tells it whether or not to load just the user's organizations, and then that parameter can be passed to the servlet. This way, this one function can handle all instances of needing to load a list of organizations to an HTML div, from anywhere in the website, as opposed to just happening when this page is loaded & having to rewrite the code for different situations.

src/main/webapp/index_script.js
andrewm898: Since ListOrganizationServlet just sends back an array of strings that each are an organization name, this function is not dealing with an object- "organization" is just one of the strings in the array.

edit: I read through the code you wrote for the popup functionality and I now understand what you mean here. ListOrg servlet is now going to return an organization object that has fields address, phone num, name, and the description.

src/main/webapp/index_script.js
Sarah-Addo: okay that makes sense!

Sarah-Addo (approved):

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -24,4 +41,61 @@
...
+    UserService userService = UserServiceFactory.getUserService();
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
+
+    Query query;
+
+    if (isUserLoggedIn && displayUserOrgs) {
+      /* If the user is logged in and wants to just see their orgs, get their user ID & index with it*/
+      String userId = userService.getCurrentUser().getUserId();
+
+      query = new Query("Distributor").setFilter(new FilterPredicate("moderatorList",
```
belraquib: This logic generally looks good, but can you check the spacing? This looks odd in github reviews. Is there a formatter you can use?

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -24,4 +41,61 @@
...
+                                                                        FilterOperator.EQUAL, userId))
+                                                                        .addSort("creationTimeStamp", SortDirection.DESCENDING);
+    } else {
+      /* If no username was included, it just returns all orgs */
+        // TODO: make this only return approved orgs
+      query = new Query("Distributor").addSort("creationTimeStamp", SortDirection.DESCENDING);
```
belraquib, line 75: This behavior (logged in vs not) would be great for a test! Suggestion: Can you pull out a function "Query ConstructQueryForUserInfo(User user)" to encapsulate the idea? This makes that unit of work testable

src/main/webapp/index_script.js
```diff
@@ -13,30 +13,22 @@
...
-    "name": "Organization E",
-  },
-  {
-    "name": "Organization F"
-  }];
   const organizationArea = document.getElementById('organization-list');
-  organizations.forEach((organization) => {
-    organizationArea.appendChild(createOrganization(organization));
-  });
+  renderOrganizations(organizationArea);
```
belraquib: instead of a function, can you create a class that
  knows how to fetch the data, storing it in a local variable
  knows how to render the stored data when it arrives?

belraquib (changes requested):

src/main/webapp/index_script.js
andrewm898: I adjusted this branch to integrate the code sarah wrote for the popups, and that code stores the organization list in a local variable after fetching it and then renders it once it arrives.

belraquib (approved):

PR #12, opened 2020-06-30
Create "Organizations" Page for Maintainer View

jnnybk:
Summary:
1) Refactor and create Organization class in JS
2) Create SearchArea element for Organization page
3) Add editing form for "Edit Organization"


src/main/webapp/organization.js
```diff
@@ -0,0 +1,233 @@
...
+
+    const organizationNameElement = document.createElement('div');
+    organizationNameElement.classList.add("organization-name");
+    organizationNameElement.textContent = this.organization.name;
+
+    this.organizationElement.addEventListener('click', () => {
+      // TODO: GetOrganizationServlet to display more information about this servlet
+
+      // Checks if organization-popup-area element exists.
+      let popupIndexPage = document.getElementById("organization-popup-area");
```
Sarah-Addo: think this should be const

src/main/webapp/style.css
```diff
@@ -1,17 +1,17 @@
-#help-near-me-content {
+.all-content {
```
Sarah-Addo, line 1: should be named something more specific like maintainer organizations content or something

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,115 @@
...
+    this.filterButtonArea.appendChild(this.filterButton);
+
+    this.organizationSearchArea.appendChild(this.filterButtonArea);
+
+    this.organizationList = document.createElement("div");
+    this.organizationList.setAttribute("id", "organization-list");
+
+    this.organizationPopupArea = document.createElement("div");
+    this.organizationPopupArea.setAttribute("id", "organization-popup-area");
+
```
Sarah-Addo, line 105: will need to add this to the hide pop up class

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,115 @@
...
+    this.filterButton.setAttribute("type", "button");
+    this.filterButton.setAttribute("class", "gray-button");
+    this.filterButton.setAttribute("id", "filter-button");
+    this.filterButton.textContent = "Filter";
+    this.filterButtonArea.appendChild(this.filterButton);
+
+    this.organizationSearchArea.appendChild(this.filterButtonArea);
+
+    this.organizationList = document.createElement("div");
+    this.organizationList.setAttribute("id", "organization-list");
```
Sarah-Addo, line 82: may need to append organizations to the list here and then append the entire list to the search area here for styling, move lines 106-111 to right after 101

src/main/webapp/edit.html
```diff
@@ -6,8 +6,23 @@
 	  <title>Givr</title>
 	  <link rel="stylesheet" href="style.css">
     <script src="user.js"></script>
+    <script src="organization.js"></script>
+    <script src="search_area.js"></script>
   </head>
 
   <body>
+    <div class="all-content">
```
Sarah-Addo, line 14: edit-content

src/main/webapp/edit.html
```diff
@@ -6,8 +6,23 @@
...
+      <div id="site-title">Givr</div>
+      <div id="nav-bar">
+        <a href="index.html">Help Near Me</a>
+        <a id="login-url">Login</a>
+      </div>
+      <div id="all-organizations">
```
Sarah-Addo, line 20: may want to remove or add todo

src/main/webapp/organization.js
```diff
@@ -0,0 +1,233 @@
...
+      if (popupIndexPage) {
+        popupIndexPage.textContent = "";
+        popupIndexPage.appendChild(this.createOrganizationPopup());
+        popupIndexPage.style.display = 'block';
```
Sarah-Addo: use class list for display

src/main/webapp/organization.js
```diff
@@ -0,0 +1,233 @@
...
+      popupElement.appendChild(popupEditButton);
+    }
+
+    const closeButtonElement = document.createElement('div');
+    closeButtonElement.classList.add("popup-close-button");
+    closeButtonElement.textContent = 'X';
+    closeButtonElement.addEventListener('click', () => {
+      // Remove the popup from the DOM.
+      if (document.getElementById("organization-popup-area")) {
+        document.getElementById("organization-popup-area").style.display = 'none';
```
Sarah-Addo: use class list for display

src/main/webapp/organization.js
```diff
@@ -0,0 +1,233 @@
...
+    const popupElement = document.createElement("div");
+    popupElement.classList.add("organization-popup");
+
+    const popupNameElement = document.createElement('div');
+    popupNameElement.classList.add("organization-name");
+    popupNameElement.textContent = this.organization.name;
+
+    const popupPhoneElement = document.createElement('div');
+    popupPhoneElement.classList.add("organization-popup-phone");
+    popupPhoneElement.textContent = this.organization.phone;
```
andrewm898: I think this field is "phoneNum" in the organization class, not phone

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,115 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+document.addEventListener("DOMContentLoaded", async function() {
+  // TODO: Fetch list of organizations using servlet.
+  let organizations = [{
```
andrewm898: Can you make it so these organizations come from a servlet call?
const response = await fetch(`/list-organizations`); <- /list-organiations is surrounded by backticks here
const organizations = await response.json();

src/main/webapp/organization.js
```diff
@@ -0,0 +1,233 @@
...
+
+    const popupAddressElement = document.createElement('div');
+    popupAddressElement.classList.add("organization-popup-address");
+    popupAddressElement.textContent = this.organization.address;
+
+    const popupEditElement = document.createElement('div');
+    if (this.isMaintainer) {
+      const popupEditButton = document.createElement('button');
```
Sarah-Addo: comment about edit button

src/main/webapp/organization.js
jnnybk: I got rid of line 42 because it would default to showing.

src/main/webapp/search_area.js
jnnybk: Should I use backticks?

jnnybk: I am assuming yes so that we can use the template literals for the params!

src/main/webapp/edit.html
Sarah-Addo: I think this is what would be in the organizations view page and not the edit html page. The way I wrote edit it that it would be a popup handled by js so I actually do not think that there is going to be a separate html page

src/main/webapp/index.html
```diff
@@ -1,26 +1,25 @@
...
-			<a id="login-url" href="">Login</a>
-      <a href="registration.html">Register Org</a> <!--this will only appear when logged in after login is setup -->
-			<a id="login-url">Login</a>
-		</div>
-		<div id="search-area">
-		</div>
-	</div>
-</body>
+  <body>
+    <div class="all-content">
```
Sarah-Addo: Could be revert it back to having a unique id? I think it would be better to have specific id names for the content of different html pages since there will probably be real differences in the near future

src/main/webapp/organization.js
```diff
@@ -0,0 +1,85 @@
...
+    this.organizationElement.classList.add("organization");
+
+    const organizationNameElement = document.createElement('div');
+    organizationNameElement.classList.add("organization-name");
+    organizationNameElement.textContent = this.organization.name;
+
+    this.organizationElement.addEventListener('click', () => {
+      // TODO: GetOrganizationServlet to display more information about this servlet
+
+      const popupIndexPage = document.getElementById("organization-popup-area");
```
Sarah-Addo: Why is it renamed popupIndexPage?

src/main/webapp/organization.js
```diff
@@ -0,0 +1,85 @@
...
+      popupIndexPage.textContent = "";
+      popupIndexPage.appendChild(this.createOrganizationPopup());
+      popupIndexPage.classList.add("show-popup")
```
Sarah-Addo: need to also remove the classList hide-pop up here. This should be the same as the pull request that was approved but not merged

src/main/webapp/organization.js
```diff
@@ -0,0 +1,85 @@
...
+    if (this.isMaintainer) {
+      //TODO: Create Edit Organization Button for Maintainer.
+    }
+
+    const closeButtonElement = document.createElement('div');
+    closeButtonElement.classList.add("popup-close-button");
+    closeButtonElement.textContent = 'X';
+    closeButtonElement.addEventListener('click', () => {
+      // Remove the popup from the DOM.
+      if (document.getElementById("organization-popup-area")) {
```
Sarah-Addo: Again I think you would just need to add and remove the show pop up class like what is currently in master. I do not think that what is currently in master should be changing in regards to popup functionality

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,94 @@
...
+  const organizations = await getListOfOrganizations();
+
+  // TODO: Get Maintainer status by checking if requester User is a Maintainer. 
+  // This checks that requester User has valid credentials to edit/delete/view ALL organizations. (by checking userID)
+  let isMaintainer = false;
+  if (document.getElementById('search-area')) {
+    const mainSearchArea = new SearchArea(document.getElementById('search-area'), organizations, isMaintainer);
+  }
+  isMaintainer = true;
+  if (document.getElementById('all-organizations')) {
```
Sarah-Addo, line 34: i cant find the all-organizations element in these files i think, would this be in the organizations.html?

src/main/webapp/style.css
Sarah-Addo: Per my first other comment would you consider having a more specific name?

Sarah-Addo (changes requested):

Sarah-Addo (approved):

src/main/webapp/organization.js
```diff
@@ -0,0 +1,85 @@
...
+    this.organizationElement.classList.add("organization");
+
+    const organizationNameElement = document.createElement('div');
+    organizationNameElement.classList.add("organization-name");
+    organizationNameElement.textContent = this.organization.name;
+
+    this.organizationElement.addEventListener('click', () => {
+      // TODO: GetOrganizationServlet to display more information about this servlet
+
+      const organizationPopupArea = document.getElementById("organization-popup-area");
```
belraquib: Instead, of this, can you consider either:
  fire a new event from the organizationElement; 
  { event = 'organization-selected' }  (you'll have to expose this.organizationElement as well)


  or pass an instance of an "OrganizationRenderer" to this class (which the search area can create and this class then can use). 

and then the search area can decide what to do / how to render the organization.

What you want to do is separate out the data from how it is rendered into the dom.

If you do the first thing, then the search area, after creating the organization can:

organization.organizationElement.attachEventListener('organization-selected', ()=> {
   const popup div = organization.createOrgPopup();
   setPopupTo(div);
});

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,94 @@
...
+  return organizations;
+}
+
+document.addEventListener("DOMContentLoaded", async function() {
+  const organizations = await getListOfOrganizations();
+
+  // TODO: Get Maintainer status by checking if requester User is a Maintainer. 
+  // This checks that requester User has valid credentials to edit/delete/view ALL organizations. (by checking userID)
+  let isMaintainer = false;
+  if (document.getElementById('search-area')) {
```
belraquib, line 30: These bits (if search-area is in the page vs if all-organizations is in the page) should be in page-specific bits of javascript.

belraquib (changes requested):

src/main/webapp/search_area.js
jnnybk: I put a TODO comment to fix this when we create separate JS files for "Help Near Me" page vs "Organizations" page. Right now we have one JS class that is used in both pages.

src/main/webapp/organization.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  createOrganization() {
+    this.organizationElement.classList.add("organization");
+
+    const organizationNameElement = document.createElement('div');
+    organizationNameElement.classList.add("organization-name");
+    organizationNameElement.textContent = this.organization.name;
+
+    // this.organizationElement.addEventListener()
```
belraquib: remove?

src/main/webapp/organization.js
```diff
@@ -0,0 +1,87 @@
...
+    const popupEditElement = document.createElement('div');
+    if (this.isMaintainer) {
+      //TODO: Create Edit Organization Button for Maintainer.
+    }
+
+    const closeButtonElement = document.createElement('div');
+    closeButtonElement.classList.add("popup-close-button");
+    closeButtonElement.textContent = 'X';
+    closeButtonElement.addEventListener('click', () => {
+      // Remove the popup from the DOM.
```
belraquib: same suggestion here; can you fire a custom "organization close" event?

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,109 @@
...
+
+    this.organizationSearchArea.appendChild(this.filterButtonArea);
+
+    this.organizationList = document.createElement("div");
+    this.organizationList.setAttribute("id", "organization-list");
+    
+    organizations.forEach((organization) => {
+      const newOrganization = new Organization(organization, isMaintainer);
+
+      newOrganization.organizationElement.addEventListener('click', () => {
```
belraquib: Move this code into Organization.createOrganization :)

src/main/webapp/search_area.js
```diff
@@ -0,0 +1,109 @@
...
+        var event = new CustomEvent('organization-selected');
+        newOrganization.organizationElement.dispatchEvent(event);
+      })
+
+      newOrganization.organizationElement.addEventListener('organization-selected', (organizationElement) => {
```
belraquib: great! keep this.

belraquib (changes requested):

src/main/webapp/organization.js
jnnybk: yes!

belraquib (approved):

PR #17, opened 2020-07-03
fix styling

Sarah-Addo:
This pull request: 
* Reorders nav bar depending on if a user is logged in and who it is
* Styles registration page 
* Styles popup and adds the rest of organization information to it
* Updates list organization servlet and organization class to include all wanted member variables

Screenshots: https://screenshot.googleplex.com/Mdpf2D7NzXx, https://screenshot.googleplex.com/DjkxCvHCXKp


src/main/webapp/registration.html
```diff
@@ -13,18 +14,39 @@
...
-      <textarea name="description" maxlength="180" placeholder="Org description"></textarea>
-      <input type="submit">
-		</form>
-  </div>
+	<div id="registration-content">
+		<div id="site-title">Givr</div>
+		<div id="registration-form-header">Register An Organization</div>
+		<div id="registration-form-area">
+			<form action="/add-organization" method="POST">
+				<label for = "registration-name" id = "registration-entry-label">Organization Name: </label>
```
andrewm898, line 22: Did you update AddOrganizationServlet to have all of these new field id's for this form?

Sarah-Addo: AddOrganizationServlet actually uses the names and not the ids when it reads in an input request so it still works the name

src/main/webapp/registration.html
```diff
@@ -4,6 +4,7 @@
 <head>
```
jnnybk, line 4: nit: can you format this file with the text formatter?

jnnybk (approved):

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -18,28 +18,39 @@
...
-  private int id;
+  private long id;
   private String name;
   private String email;
   private String address;
   private String description;
   private String phoneNum;
-  private int openHour;
-  private int closingHour;
+  private long openingHour;
```
mickrobk, line 27: Is this openingHour or openingTimestamp? (Same for closingHour)

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -81,14 +81,23 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     ArrayList<Organization> requestedOrganizations = new ArrayList<Organization>();
 
-    /* Fills requestedOrganizations array with 4 fields of each org- name, phone, addres, and desc. */
+    /* Fills requestedOrganizations array*/
     for (Entity entity : results) {
-
-      Organization newOrg = new Organization((String) entity.getProperty("orgName"),
-                                             (String) entity.getProperty("orgEmail"),
+      long openingHour = (long) ((ArrayList)entity.getProperty("openHours")).get(0);
+      long closingHour = (long) ((ArrayList)entity.getProperty("openHours")).get(1);
```
mickrobk, line 87: Is there a richer schema we can use for this? It feels error prone to have a convention that the second entry in "openHours" is the closing timestamp.

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -81,14 +81,23 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      Organization newOrg = new Organization((long) entity.getKey().getId(),
```
mickrobk, line 88: This function is getting quite long. How would you start to factor it apart into reusable components?

What do you think of moving this functionality to the Organizations class?

src/main/webapp/user.js
```diff
@@ -28,26 +28,62 @@ class User {
 
     if (loginData.isLoggedIn) {
       this.isMaintainer = loginData.user.isMaintainer;
-      document.getElementById('login-url').innerText = "Log Out";
-      document.getElementById('login-url').href = loginData.url;
 
+      // TODO: set maintainer status correctly, manually doing this for testing purposes
```
mickrobk, line 32: This is a reasonable way to make progress now, but we need to be extremely careful that we don't leave this here, as it'd be a serious security bug. Instead of just writing "TODO" can you write: "TODO(<issue number>)" and create a high priority github issue to fix this?

Same for the other TODO below

mickrobk (changes requested):
This PR looks like it does a few major things:
 1) Update the Java organization class
 2) Update the navbar
 3) Update the register org page
 4) Update the css

I left comments for (1), but didn't review 2,3,4. In general we should aim for 1 change per PR, as it's easier to review, rollback/reason about project history, and easier to collaborate on. Can you split this PR into individual chunks?

PR #20, opened 2020-07-06
update to use history entities instead of strings

andrewm898:
This updates the timestamping to use seconds since epoch time, and changes the history arraylist to use embedded entities instead of storing a string. 

note: I know we originally intended to use a custom class for this; it's much simpler to just use an embedded entity & you get exactly the same functionality.


jnnybk (approved):

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -39,8 +45,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String hourOpen = request.getParameter("hour-open");
     String hourClosed = request.getParameter("hour-closed");
 
-    //TODO: get timestamp with transactions instead
-    long timestampMillis = System.currentTimeMillis();
+    long secondSinceEpoch = Instant.now().toEpochMilli();
```
Sarah-Addo: seconds since epoch or milliseconds. Could you also add a comment here since I dont understand what this does

Sarah-Addo (approved):

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +60,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    ArrayList changeHistory = new ArrayList<>();
+    UserService userService = UserServiceFactory.getUserService();
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
+    String username;
+    if (isUserLoggedIn) {
+      /* Currently uses user email to be consistent w other parts of codebase, subject to change */
+      username = userService.getCurrentUser().getEmail();
+    } else {
+      username = "Unknown Author";
+    }
```
mickrobk: Are anonymous users allowed to create organizations? How did we get here?

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +60,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    moderatorList.add("anon creator"); //dummy value for before auth gets set up
+    /* This implementation stores history entries as embedded entities instead of custom objects
+     * because it is much simpler that way */
+    ArrayList changeHistory = new ArrayList<>();
+    UserService userService = UserServiceFactory.getUserService();
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
+    String username;
+    if (isUserLoggedIn) {
+      /* Currently uses user email to be consistent w other parts of codebase, subject to change */
+      username = userService.getCurrentUser().getEmail();
```
mickrobk: Is there a issue tracking this? Stylistically, it'd be better to add a TODO(<issue number>): description so that this type of debt can be tracked and fixed.

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +60,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    ArrayList<String> moderatorList = new ArrayList<String>();
-    moderatorList.add("anon creator"); //dummy value for before auth gets set up
+    /* This implementation stores history entries as embedded entities instead of custom objects
+     * because it is much simpler that way */
+    ArrayList changeHistory = new ArrayList<>();
+    UserService userService = UserServiceFactory.getUserService();
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
+    String username;
+    if (isUserLoggedIn) {
+      /* Currently uses user email to be consistent w other parts of codebase, subject to change */
```
mickrobk: Normally I'm all for consistency, but when I'd be extra careful about the cost of a bug in history tracking. What's the downside of moving to the correct id system now?

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +60,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    } else {
+      username = "Unknown Author";
+    }
+    
+    EmbeddedEntity historyEntry = new EmbeddedEntity();
+
+    historyEntry.setProperty("changeAuthor", username);
+    historyEntry.setProperty("changeMessage", "Organization was registered");
+    historyEntry.setProperty("changeTimeStamp", millisecondSinceEpoch);
+    changeHistory.add(historyEntry);
```
mickrobk: How will this scale to other types of history additions? Is there functionality here that should be factored out?

mickrobk (changes requested):

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: No, this situation should never happen, I just added this so if it does happen or for whatever reason the userservice doesn't pick up an email it's flagged as an unknown author

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: I changed it to email because the current other PR's are referencing users by their email, so I designed this system to reference them in the same way

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
mickrobk: In general it'll be easier to make the service robust if you fail on conditions that shouldn't happen instead of accepting them. Either way you're being defensive, but by failing you:
 1) Get alerted to the fact that something is broken
 2) Stay more secure. (Eg a bug doesn't enable unauthorized users access to org creation)
 3) Write less code. (Preconditions are simpler than error handling)

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: I wrote that it's subject to change if we decide to change to tracking users by userId instead of email, but right now it's using the emails to ensure that the history entries can show who wrote it without needing to do a conversion on the userid property

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
mickrobk: Right, but should other parts of the codebase be doing this? What's the plan to fix it, and how does this new code fit into that plan?

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: I figured for the functionalities we have right now it makes sense to record a message, author, and timestamp, and for more complicated tasks such as editing a moderator list we can either change the history entities to have more fields, or just append together a string message for it & follow this same format. For now, I think the strategy of just putting together a message for each change will scale to other additions fine though.

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +60,31 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
-    ArrayList<String> moderatorList = new ArrayList<String>();
-    moderatorList.add("anon creator"); //dummy value for before auth gets set up
+    /* This implementation stores history entries as embedded entities instead of custom objects
+     * because it is much simpler that way */
+    ArrayList changeHistory = new ArrayList<>();
+    UserService userService = UserServiceFactory.getUserService();
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
+    String username;
+    if (isUserLoggedIn) {
```
belraquib: Should you allow an organization edit if the user is not logged in?

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
belraquib: Though the other parts of the code will change to pivot around userId, I think it's OK to break consistency here as long as you are tracking the change elsewhere.

belraquib (changes requested):

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: Can we talk about userId vs email as a pod later today? Jenny Sarah and I were thinking that for an MVP it would make sense to just reference users by full email address since that can't be the same for multiple different users, but I know there are some drawbacks related to that. I think for this feature though I'm going to change it to record the userId and the email.

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: My first implementation just allowed this, but this new commit throws an IllegalArgumentException instead to protect the database from unknown users

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: That makes sense- the new code throws an exception instead

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
mickrobk: What if someone other than you creates those pages and names the properties slightly differently? 

Looking at these fields, is there a common core that could be pulled out and work with where we're planning to go? (AKA how could you generalize this. Referring back to the design docs could be helpful here)

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
mickrobk: If the entire function has the same access level, it'd be better to perform the access check at the start of the function, s.t. as the function evolves over time it stays secure. Security here means items ranging from the obvious: another transaction to more subtle: like timing attacks

mickrobk (changes requested):

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: In our design doc we wanted to generalize this by creating a history object and writing that into the datastore, but I found that it's much simpler to use an embedded entity than set up a customized type like that, so we decided to change the history structure to that and agree on using these property names for all history writes

mickrobk: I get that it's less code in this PR to add this feature inline, but I'm not convinced it's simpler for the project as a whole. :) Agreeing on names etc is a lot more error prone than using the compiler / a type system, and it seems like you (or someone else in the pod) will immediately need to redo and generalize this once another mutating doFn is added.

Could you take a stab at a more general solution?

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +72,23 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     openHours.add(Integer.parseInt(hourOpen));
     openHours.add(Integer.parseInt(hourClosed));
 
-    ArrayList<String> moderatorList = new ArrayList<String>();
-    moderatorList.add("anon creator"); //dummy value for before auth gets set up
+    /* This implementation stores history entries as embedded entities instead of custom objects
+     * because it is much simpler that way */
+    ArrayList changeHistory = new ArrayList<>();
+    
+    EmbeddedEntity historyEntry = new EmbeddedEntity();
```
belraquib: May I suggest extracting a function that you can use later:

class HistoryManager {
  void PrepareForOrganization(Entity organization);
  void RecordHistory(userId, message, timestamp);
}

My imagined use case would be:

// This will read the current history and set up its own state if one exists. It also calls
// newOrganization.setProperty("changeHistory", ...); to point to local memory
HistoryManager history = new HistoryManager(newOrganization);

// This creates the embedded entity and adds it to the history
history.RecordHistory(userId, "Org was registered", millisecondsSinceEpoch);

newOrganization.setProperty(...); // all other changes

That way you can use the history manager later as well.

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -52,15 +72,23 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+
+    historyEntry.setProperty("changeAuthorEmail", username);
+    historyEntry.setProperty("changeAuthorId", userId);
+    historyEntry.setProperty("changeMessage", "Organization was registered");
+    historyEntry.setProperty("changeTimeStamp", millisecondSinceEpoch);
```
belraquib: Remember to reflect the units in the variable name; changeTimestampMillis

belraquib (changes requested):

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: I agree- I'm working on following Tay's advice below about making a HistoryManager class to take care of these things wherever it is called

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
andrewm898: I wrote this class to just have the function recordHistory, which takes in the parameters and then returns an entity that can be added to the history arraylist from wherever it is being called.

src/main/java/com/google/sps/data/HistoryManager.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EmbeddedEntity;
+
+
+
+public final class HistoryManager {
+
+  public HistoryManager() {
```
jnnybk: do we need the default constructor if it is doing nothing?

jnnybk (approved):

src/main/java/com/google/sps/data/HistoryManager.java
andrewm898: good catch- I removed that

src/main/java/com/google/sps/data/HistoryManager.java
```diff
@@ -0,0 +1,35 @@
...
+
+
+public final class HistoryManager {
+
+  public EmbeddedEntity recordHistory(String userId, String historyMessage, long millisecondSinceEpoch) {
+    EmbeddedEntity historyEntry = new EmbeddedEntity();
+
+    historyEntry.setProperty("changeAuthorId", userId);
+    historyEntry.setProperty("changeMessage", historyMessage);
+    historyEntry.setProperty("changeTimeStamp", millisecondSinceEpoch);
```
belraquib: this needs a unit

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -17,18 +17,37 @@
...
 import com.google.sps.data.User;
+import com.google.sps.data.HistoryManager;
+import java.time.Instant;
 import java.io.IOException;
 
 @WebServlet("/add-organization")
 public class AddOrganizationServlet extends HttpServlet {
 
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    UserService userService = UserServiceFactory.getUserService();
```
belraquib, line 39: Once jenny's change is in, you can use that!

src/main/java/com/google/sps/data/HistoryManager.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.appengine.api.datastore.EmbeddedEntity;
+
+
+
+public final class HistoryManager {
+
+  public EmbeddedEntity recordHistory(String userId, String historyMessage, long millisecondSinceEpoch) {
+    EmbeddedEntity historyEntry = new EmbeddedEntity();
+
+    historyEntry.setProperty("changeAuthorId", userId);
```
belraquib, line 33: please consider a separate PR to change these (the property strings) to use constants.

This makes the error a compile time error!!

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -37,21 +56,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    // TODO change this to our History object- for prototyping just using strings
-    ArrayList<String> changeHistory = new ArrayList<String>();
-    changeHistory.add("Organization was registered");
 
-    newOrganization.setProperty("creationTimeStamp", timestampMillis);
-    newOrganization.setProperty("lastEditTimeStamp", timestampMillis);
+    ArrayList<String> moderatorList = new ArrayList<String>();
+    moderatorList.add(username);
+
+    newOrganization.setProperty("creationTimeStampMillis", millisecondSinceEpoch);
```
belraquib, line 81: Leave a TODO to replace 80-84 with sarah's changes to update the organization from a request.

belraquib (changes requested):

src/main/java/com/google/sps/data/HistoryManager.java
andrewm898: I added this to the issues list

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -37,21 +56,30 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    changeHistory.add(history.recordHistory(userId, "Organization was registered", millisecondSinceEpoch));
+    newOrganization.setProperty("changeHistory", changeHistory);
 
-    // TODO change this to our History object- for prototyping just using strings
-    ArrayList<String> changeHistory = new ArrayList<String>();
-    changeHistory.add("Organization was registered");
 
-    newOrganization.setProperty("creationTimeStamp", timestampMillis);
-    newOrganization.setProperty("lastEditTimeStamp", timestampMillis);
+    ArrayList<String> moderatorList = new ArrayList<String>();
```
belraquib, line 77: Add a TODO; this should eventually be userids, right?

belraquib (approved):

PR #21, opened 2020-07-06
update organization class

Sarah-Addo:


src/main/java/com/google/sps/data/Organization.java
```diff
@@ -18,28 +18,39 @@
...
-  private int id;
+  private long id;
   private String name;
   private String email;
   private String address;
   private String description;
   private String phoneNum;
-  private int openHour;
-  private int closingHour;
+  private long openingHour;
```
Sarah-Addo: robs comment: Is this openingHour or openingTimestamp? (Same for closingHour)

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -81,14 +81,23 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     QueryResultList<Entity> results = prepQuery.asQueryResultList(fetchOptions);
     ArrayList<Organization> requestedOrganizations = new ArrayList<Organization>();
 
-    /* Fills requestedOrganizations array with 4 fields of each org- name, phone, addres, and desc. */
+    /* Fills requestedOrganizations array*/
     for (Entity entity : results) {
-
-      Organization newOrg = new Organization((String) entity.getProperty("orgName"),
-                                             (String) entity.getProperty("orgEmail"),
+      long openingHour = (long) ((ArrayList)entity.getProperty("openHours")).get(0);
```
Sarah-Addo: robs comment: Is there a richer schema we can use for this? It feels error prone to have a convention that the second entry in "openHours" is the closing timestamp.

Sarah-Addo: robs comment: Is there a richer schema we can use for this? It feels error prone to have a convention that the second entry in "openHours" is the closing timestamp.

andrewm898: We're planning on making a more detailed data structure for this that shows all of the open / closed times for each individual day of the week, but that is still a TODO()

src/main/java/com/google/sps/data/Organization.java
Sarah-Addo: openingHour which is for the organizations opening times. Datastore stores these as longs. Should I change it to int and convert it where needed?

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
Sarah-Addo: Agreed with andrews comment. Will add a TODO()

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -81,14 +81,25 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    /* Fills requestedOrganizations array*/
     for (Entity entity : results) {
-
-      Organization newOrg = new Organization((String) entity.getProperty("orgName"),
-                                             (String) entity.getProperty("orgEmail"),
+      
+      // TODO(): Implement better schema to represent opening and closing hours for different days
+      long openingHour = (long) ((ArrayList)entity.getProperty("openHours")).get(0);
+      long closingHour = (long) ((ArrayList)entity.getProperty("openHours")).get(1);
+      Organization newOrg = new Organization((long) entity.getKey().getId(),
```
Sarah-Addo: robs comment: This function is getting quite long. How would you start to factor it apart into reusable components?

What do you think of moving this functionality to the Organizations class?

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
mickrobk: As a question for the team: if we know we have to redo this part of the schema, what's the advantage of checking in this implementation vs leaving it out entirely as a simplification?

src/main/java/com/google/sps/data/Organization.java
mickrobk: Breaking down the question a bit:
 1) What's currently being stored?
 2) Does the current name reflect what's being stored?
 3) What should be stored?

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: We looked at this function and felt that it makes sense to just use the constructor for the organizations class, and that the other tasks are pretty specific to what this servlet does. What are your ideas for ways we can break this servlet up into reusable parts? @mickrobk

mickrobk: What are the alternatives you've considered?

Sarah-Addo: We considered initializing these objects when we do the Add Organization servlet but we figured that it was be easier to initialize them here since we are getting the entities that we need at this moment rather than searching for the organization object that the entity corresponds to.

We are now considering creating a setter in the organizations class that takes in an entity and sets the member variables according to the entity properties

andrewm898: We thought about alternatives some more and realized that we could simplify this function a lot if we create a setter in the Organization class that takes in an entity object and sets all of its properties to the organizations properties. This could be used anywhere we want to populate the fields of an organization object with an entity

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
Sarah-Addo: We recognize that its an important aspect of the overall project (people being able to see when an organization is opened or closed) which is why we wanted to include a simpler implementation before we dive into a more complex implementation

src/main/java/com/google/sps/data/Organization.java
Sarah-Addo: 1. Currently the organization can input the hour that there organization opens and closes which is being stored in the respective variables
2. I do believe so since for now we are only allowing them to input a general opening hour and a general closing hour. What did you have in mind?
3. In my mind we are storing organization information that will be displayed to the user (a single number and not a timestamp). Are you talking about int vs long or a timestamp?

mickrobk: """In my mind we are storing organization information that will be displayed to the user (a single number and not a timestamp). Are you talking about int vs long or a timestamp?"""

This is exactly what I'm trying to get at. To me as a reader this looked like a unix timestamp (hence a long). If it's an hour of the day: long doesn't make much sense to me.

Int however is also underspecified. Is 11pm ==23? How do you represent the time zone?

Sarah-Addo: will go through and deprecate openinghour and closinghour in another PR that will address whatever files use these variables

belraquib (approved):

PR #22, opened 2020-07-06
add registration page

### LINKED ISSUE #35: Handle form submissions that have null value protection

Sarah-Addo:


andrewm898:
Since this PR changes the field ID's for the form that interacts with add organization servlet, those parameter names should be updated in add organization servlet in this PR

src/main/webapp/registration.html
```diff
@@ -13,18 +14,39 @@
...
-  </div>
+	<div id="registration-content">
+		<div id="site-title">Givr</div>
+		<div id="registration-form-header">Register An Organization</div>
+		<div id="registration-form-area">
+			<form action="/add-organization" method="POST">
+				<label for = "registration-name" id = "registration-entry-label">Organization Name: </label>
+				<input id = "registration-name" class = "registration-entry" type = "text" name = "org-name"/>
+
+				<label for = "registration-email" id = "registration-entry-label">Primary Email: </label>
```
mickrobk: Do you need ids for these labels? (And why are all the id's the same?)

src/main/webapp/registration.html
```diff
@@ -13,18 +14,39 @@
...
+				<label for = "registration-hour-closed" id = "registration-entry-label">Hour Organization Closes: </label>
+				<input id = "registration-hour-closed"  class = "registration-entry" type = "number" min ="0" max = "23" name = "hour-closed"/>
+
+				<label for = "registration-url-link" id = "registration-entry-label">Organization Website Link: </label>
+				<input id = "registration-url-link"  class = "registration-entry" type = "text" name = "url-link"/>
+
+				<label for = "registration-description" id = "registration-entry-label">Organization Description: </label>
+				<textarea id = "registration-description"  class = "registration-entry" name ="description" maxlength = "180"></textarea>
+
+				<input id = "registration-submit-button" class = "gray-button" type="submit"/>
```
mickrobk: I'd avoid calling a style "gray", and aim for semantic meaning instead. (What if you want to change the color of all the submit buttons to blue?)

mickrobk (changes requested):

src/main/webapp/registration.html
Sarah-Addo: No, will be removing!

andrewm898:
It looks like this has merge conflicts with the old way of updating the parameters- could you resolve it to use the new code in this PR?

src/main/java/com/google/sps/servlets/AddOrganizationServlet.java
```diff
@@ -17,65 +17,77 @@
...
-    openHours.add(Integer.parseInt(hourClosed));
+    changeHistory.add(history.recordHistory("Organization was registered", millisecondSinceEpoch));
+    newOrganizationEntity.setProperty("changeHistory", changeHistory);
 
+    //TODO use UserId's here
     ArrayList<String> moderatorList = new ArrayList<String>();
-    moderatorList.add("anon creator"); //dummy value for before auth gets set up
+    moderatorList.add(user.getUserId());
+
+    newOrganizationEntity.setProperty("creationTimeStampMillis", millisecondSinceEpoch);
```
belraquib, line 69: Leave a comment as to why you are setting these directly here instead of organization updater?

belraquib (approved):

PR #23, opened 2020-07-06
rebuild navbar depending on user login status

Sarah-Addo:


src/main/webapp/user.js
```diff
@@ -28,21 +28,56 @@ class User {
 
     if (loginData.isLoggedIn) {
       this.isMaintainer = loginData.user.isMaintainer;
-      document.getElementById('login-url').innerText = "Log Out";
-      document.getElementById('login-url').href = loginData.url;
 
+      // TODO: set maintainer status correctly, manually doing this for testing purposes
```
Sarah-Addo: robs comment: This is a reasonable way to make progress now, but we need to be extremely careful that we don't leave this here, as it'd be a serious security bug. Instead of just writing "TODO" can you write: "TODO()" and create a high priority github issue to fix this?

Same for the other TODO below

belraquib (approved):

PR #25, opened 2020-07-06
add edit org functionality and style

### LINKED ISSUE #35: Handle form submissions that have null value protection

Sarah-Addo:
Creates the edit organization page and does the servlet action.
Screenshot: https://screenshot.googleplex.com/VwM6Q0CwKz0

### DIFF
--- a/src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
+++ b/src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
@@ -32,6 +32,7 @@ import java.text.SimpleDateFormat;
 import com.google.sps.data.User;
 import com.google.sps.data.HistoryManager;
 import com.google.sps.data.OrganizationUpdater;
+import com.google.sps.data.GivrUser;
 import java.time.Instant;
 import java.io.IOException;
 import com.google.gson.Gson;
@@ -54,23 +55,16 @@ public class EditOrganizationServlet extends HttpServlet {
         response.sendError(HttpServletResponse.SC_NOT_FOUND);
         return;
     }
-
+    
+    GivrUser user = GivrUser.getLoggedInUser();
     OrganizationUpdater organizationUpdater = new OrganizationUpdater(organizationEntity);
     
-    if(isMaintainer) {
-      try {
-        organizationUpdater.updateOrganization(request, /*isMaintainer*/ true, /*isModerator*/ false);
-      } catch(IllegalArgumentException err) {
-          response.sendError(HttpServletResponse.SC_NOT_FOUND);
-          return;
+    try {
+      organizationUpdater.updateOrganization(request, user, /*forRegistration*/ false);
+    } catch(IllegalArgumentException err) {
+        response.sendError(HttpServletResponse.SC_NOT_FOUND);
+        return;
       }
-    } else {
-        try {
-          organizationUpdater.updateOrganization(request, /*isMaintainer*/ false, /*isModerator*/ true);
-        } catch(IllegalArgumentException err) {
-            response.sendError(HttpServletResponse.SC_NOT_FOUND);
-            return;
-        }    
     }
 
     // updates entity with changed properties from the form
--- a/src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
+++ b/src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
@@ -24,8 +24,11 @@ import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.Entity;
 import java.util.ArrayList;
+import java.util.Map;
+import java.util.HashMap;
 import java.util.Arrays;
 import com.google.sps.data.User;
+import com.google.sps.data.Organization;
 import java.io.IOException;
 import com.google.gson.Gson;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,85 @@
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    long id = Long.parseLong(request.getParameter("id"));
+    Key organizationKey = KeyFactory.createKey("Distributor", id);
+    Entity organization = new Entity("Distributor");
+
+    // try catch for compliation purposes, servlet will not be called without a valid id param
+    try {
+      organization = datastore.get(organizationKey);
+    } catch(com.google.appengine.api.datastore.EntityNotFoundException err) {
+        System.out.println("Entity Not Found");
```
belraquib: return early here with a 404 response?

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,85 @@
...
+    String newOrgEmail = request.getParameter("email");
+    String newOrgStreetAddress = request.getParameter("address");
+    String newOrgPhoneNum = request.getParameter("phone-number");
+    String newOrgUrl = request.getParameter("url-link");
+    String newOrgDescription = request.getParameter("description");
+    String newHourOpen = request.getParameter("hour-open");
+    String newHourClosed = request.getParameter("hour-closed");
+    String newApproval = request.getParameter("approval");
+    String newModeratorList = request.getParameter("moderator-list");
+
```
belraquib: check that all the above are not-null?

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,85 @@
...
+    // try catch for compliation purposes, servlet will not be called without a valid id param
+    try {
+      organization = datastore.get(organizationKey);
+    } catch(com.google.appengine.api.datastore.EntityNotFoundException err) {
+        System.out.println("Entity Not Found");
+    }
+
+    // Get all information from form
+    String newOrgName = request.getParameter("org-name");
```
belraquib: could you create a map of param -> property name?

"org-name" -> "orgName", etc

then you can write this as a loop! (at least for the properties you copy over)

for formKey, objKey:
   String value = request.getParameter(formKey);
   if (value is null) fail
   organization.setProperty(objKey, value);

Bonus points if you can figure out how to coerce "approval" into true/false in the above scheme (hint; you don't want to just store the object Key in the map)

belraquib (changes requested):

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,68 @@
...
 public class EditOrganizationServlet extends HttpServlet {
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
+    //Get the proper organization entity for updating
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    long organizationId = Long.parseLong(request.getParameter("id"));
+    Key organizationKey = KeyFactory.createKey("Distributor", organizationId);
+    Entity organizationEntity = new Entity("Distributor");
+
+    // try catch for compliation purposes, servlet will not be called without a valid id param
```
belraquib: compilation*

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,68 @@
...
 
 @WebServlet("/edit-organization")
 public class EditOrganizationServlet extends HttpServlet {
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
 
+    //Get the proper organization entity for updating
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    long organizationId = Long.parseLong(request.getParameter("id"));
+    Key organizationKey = KeyFactory.createKey("Distributor", organizationId);
+    Entity organizationEntity = new Entity("Distributor");
```
belraquib: start this off as null? you are overwriting it in 52 anyway.

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,68 @@
...
+
+    // try catch for compliation purposes, servlet will not be called without a valid id param
+    try {
+      organizationEntity = datastore.get(organizationKey);
+    } catch(com.google.appengine.api.datastore.EntityNotFoundException err) {
+        response.sendError(HttpServletResponse.SC_NOT_FOUND);
+        return;
+    }
+    
+    GivrUser user = GivrUser.getLoggedInUser();
```
belraquib, line 58: should you check if the user is logged in here and throw an error otherwise?

belraquib: Seems like a good function to add to GivrUser:

void ThrowIfNotLoggedIn() {
}

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
```diff
@@ -18,10 +18,68 @@
...
+    }
+
+    // updates entity with changed properties from the form
+    organizationEntity = organizationUpdater.getEntity();
+
+    //TODO: get timestamp with transactions instead
+    long millisecondSinceEpoch = Instant.now().toEpochMilli();
+    organizationEntity.setProperty("lastEditTimeStampMillis", millisecondSinceEpoch);
+
+    ArrayList<EmbeddedEntity> changeHistory = (ArrayList) organizationEntity.getProperty("changeHistory");
```
belraquib, line 76: Suggestion: if updateOrganization accepted the result from history.recordHistory() you can keep all the update logic in OrganizationUpdater. Thoughts?

belraquib (changes requested):

src/main/webapp/index_script.js
```diff
@@ -72,15 +74,16 @@ class SearchArea {
...
+    * loads by not passing any parameters to /list-organizations
+    */
   async requestAndDisplayOrganizations() {
+    //TODO: fetch organizations with param
     const response = await fetch(`/list-organizations`);
-    const organizationNames = await response.json();
-    for (let i = 0; i < organizationNames.length; i++) {
-      this.organizationList.appendChild(this.createOrganization(organizationNames[i]));
+    const organizations = await response.json();
+    for (let i = 0; i < organizations.length; i++) {
```
belraquib: for (const org of organizations) {
   this.organziationList.appendChild(org);
}

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
Sarah-Addo: Currently this check happens a bit earlier since the edit button will only be shown for logged in users but will put that function on the list of refactoring for GivrUser

src/main/java/com/google/sps/servlets/EditOrganizationServlet.java
Sarah-Addo: The way I envisioned the Updater was that it was primarily parse form input from the user. We could definitely move all the update logic to the end of it but I think we should essentially move lines 73-79 to the end of it as well as modify it to handle what the history would be when addorganizationservlet calls it as well as set the creation timestamp. Your thoughts?

belraquib: It both parses and stores all the changes you want on the entity, right? Seems like 73-79 would be done in both edit and add org (and also in update good, etc). Doing it in another PR sounds fine, but file an issue to track it.

belraquib (approved):

PR #27, opened 2020-07-07
Function setup

andrewm898:
Sets up repeatable helper functions with TODO tasks contained to make addressing them in the future easier

Functions set up (all in ListOrg servlet):
1. userIsMaintainer- checks if user is maintainer in the datastore
2. getQueryFromParams- all of the things this servlet was previously doing to set up the query were moved here, and to set up filtering we just need to complete the TODO in this function that will read through the rest of the request parameters

ConstructQueryForUserInfo was removed because this functionality is now inline in getQueryFromParams function


jnnybk (approved):
LGTM! PR #28 implements the userIsMaintainer :)

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -96,9 +73,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    /* displayUserOrgsParameter is true when user only wants to see orgs they moderate*/
+    String displayUserOrgsParameter = request.getParameter("displayUserOrgs");
+    boolean displayUserOrgs = false;
+    if ((displayUserOrgsParameter != null) && (displayUserOrgsParameter.equals("true"))) {
+      /* If a parameter was sent & is set to 'true', then the displayUserOrgs boolean changes to true */
+      displayUserOrgs = true;
+    }
+
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
```
jnnybk, line 85: #28 would allow us to just call User.getLoggedInUser()!

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -96,9 +73,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    String userId = userService.getCurrentUser().getUserId();
+    boolean userIsMaintainer = userIsMaintainer(userId);
+
+    if (isUserLoggedIn && displayUserOrgs) {
+      /* If the user is logged in and wants to just see their orgs, get their user ID & index with it*/
+      query.setFilter(new FilterPredicate("moderatorList", FilterOperator.EQUAL, userId));
```
belraquib, line 91: Add a TODO: 

can you create a new function in the User class that returns the filter query? (maybe not needed?)

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -96,9 +73,39 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  public boolean userIsMaintainer(String userId) {
+    //TODO(): Check if given user ID has maintainer role in the datastore
+    return true;
+  }
+
+  /* This function constructs a query based on the request parameters & user's role */
+  public Query getQueryFromParams(HttpServletRequest request, UserService userService) {
+    Query query = new Query("Distributor").addSort("creationTimeStamp", SortDirection.DESCENDING);;
+
+    /* displayUserOrgsParameter is true when user only wants to see orgs they moderate*/
```
belraquib, line 82: Can you extract a function and create a new class:

RequestParameterHelper {
  bool getParameterAsBooleanOrDefault(String param, bool default) {
     String value = request.getParam(..);
     if (value == null) return default;
     // return if true.
  }
}

belraquib (changes requested):

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: I set up a prototype for this TODO in the User class, and then wrote a detailed Github issue about the how/why/what for the filtering backend that addresses what needs to be done with these TODO's

src/main/java/com/google/sps/data/User.java
```diff
@@ -22,4 +27,12 @@ public User(String id, boolean isMaintainer) {
     this.id = id;
     this.isMaintainer = isMaintainer;
   }
+
+  public Filter getUserFilter() {
```
belraquib: This is a userFilter against a the user table right? 

getUserFilterForUserTable?

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -92,10 +71,47 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    // TODO(): Read through request parameters and for all valid parameters and use them to modify query (filtering)
 
-  public Query ConstructQueryForUserInfo(String userID) {
-    Query query = new Query("Distributor").setFilter(new FilterPredicate("moderatorList",
-                    FilterOperator.EQUAL, userID)).addSort("creationTimeStampMillis", SortDirection.DESCENDING);
     return query;
   }
+
+  public boolean userOrgsParameterHelper(HttpServletRequest request) {
```
belraquib: suggestion: pass in a String key here.

now the function can be called:

coerceParameterToBoolean(request, key) {
  // all your code is the same, but instead of getParameter("displayUserOrgs") you do getParameter(key)
}

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -92,10 +71,47 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+     /* displayUserOrgsParameter is true when user only wants to see orgs they moderate*/
+    String displayUserOrgsParameter = request.getParameter("displayUserOrgs");
+
+    if ((displayUserOrgsParameter != null) && (displayUserOrgsParameter.equals("true"))) {
```
belraquib: This is one of my pet peeves; consider what this is saying:

if (condition evaluates to true) 
  return true;
else
  return false;

Can you consider a simpler way to phrase it

belraquib (changes requested):

src/main/java/com/google/sps/data/User.java
andrewm898: I was thinking this could operate for a single user that is requesting their orgs to be listed, not the table of all users

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: Yes!- I changed it to this in the new commit :)

belraquib (approved):

PR #28, opened 2020-07-07
Add user functionality for get logged in user and user from datastore

### LINKED ISSUE #29: Possibly remove AddUserServlet
Discuss if AddUserServlet is needed: Should any User be added? AddMaintainerServlet handles checking maintainer status and also returns the current User's object.

### LINKED ISSUE #34: Support OAuth
If time permits, integrate OAuth and replace Google Authentication API.

jnnybk:
Add functionality for User class to access current user and their information. This will allow other Java servlets to access this information without having to re-write existing code.


src/main/java/com/google/sps/data/User.java
```diff
@@ -14,12 +14,62 @@
...
+    PreparedQuery preparedQuery = datastore.prepare(query);
+    QueryResultList<Entity> userResult = preparedQuery.asQueryResultList(fetchOptions);
+
+    boolean isMaintainer = false;
+    String userEmail = "";
+    for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
+      isMaintainer = (boolean) entity.getProperty("isMaintainer");
+      userEmail = (String) entity.getProperty("userEmail");
+    }
+    user = new User(userId, isMaintainer, userEmail);
```
andrewm898: Does this check for / handle cases where that user id isn't found?

jnnybk: Added! Thanks!

src/main/java/com/google/sps/data/User.java
```diff
@@ -14,12 +14,66 @@
...
+    boolean isMaintainer = false;
+    String userEmail = "";
+
+    if (userResult.size() < 1) {
+      return null;
+    }
+    
+    for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
+      isMaintainer = (boolean) entity.getProperty("isMaintainer");
+      userEmail = (String) entity.getProperty("userEmail");
```
belraquib: I don't think we should be storing the email. You should read it from the user service.

src/main/java/com/google/sps/data/User.java
```diff
@@ -14,12 +14,66 @@
...
+  private String userEmail;
 
-  public User(String id, boolean isMaintainer) {
+  public User(String id, boolean isMaintainer, String userEmail) {
     this.id = id;
     this.isMaintainer = isMaintainer;
+    this.userEmail = userEmail;
+  }
+
+  public boolean getMaintainerStatus() {
```
mickrobk: nit: Is there a different name you could use for this? I'm picturing a hypothetical use case:

myFn {
  if (user.getMaintainerStatus()) {
     // Is the user a maintainer, or was I able to successfully query the db / find out if they were logged in?
  }
}

What you you think about naming this "IsMaintainer()"?

src/main/java/com/google/sps/data/User.java
mickrobk: Do we need to store the user email? Per the pod conv, can we find this from the auth service

src/main/java/com/google/sps/data/User.java
```diff
@@ -14,12 +14,66 @@
...
+    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(1);
+    PreparedQuery preparedQuery = datastore.prepare(query);
+    QueryResultList<Entity> userResult = preparedQuery.asQueryResultList(fetchOptions);
+
+    boolean isMaintainer = false;
+    String userEmail = "";
+
+    if (userResult.size() < 1) {
+      return null;
+    }
```
mickrobk: Do you have enough information to still return a valid user here? (Eg if the query failed, you know they aren't a maintainer)

mickrobk (changes requested):

src/main/java/com/google/sps/data/User.java
```diff
@@ -14,12 +14,66 @@
...
+    
+    for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
+      isMaintainer = (boolean) entity.getProperty("isMaintainer");
+      userEmail = (String) entity.getProperty("userEmail");
+    }
+    User user = new User(userId, isMaintainer, userEmail);
+    return user;
+  }
+
+  public static User getLoggedInUser() {
```
belraquib: You also probably want:

static User getUserFromEmail(String email) {
You should be able to look up the user id with:

https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User.html


}

src/main/java/com/google/sps/data/User.java
```diff
@@ -14,12 +14,66 @@
...
+    UserService userService = UserServiceFactory.getUserService();
+    boolean isUserLoggedIn = userService.isUserLoggedIn();
+    
+    if (isUserLoggedIn) {
+      return getUserFromDatastore(userService.getCurrentUser().getUserId());
+    } else {
```
belraquib: style: remove the else; the else


if (isUserLoggedIn) {
   return ...;
}
return null;

belraquib (changes requested):

src/main/java/com/google/sps/data/User.java
jnnybk: Hm, I'm not sure if I'm understanding this comment. The maintainer status doesn't really matter because this method is just getting a user with a string ID.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -0,0 +1,81 @@
...
+    Filter queryFilter = new FilterPredicate("userId", FilterOperator.EQUAL, userId);
+    Query query = new Query("User").setFilter(queryFilter);
+      
+    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(1);
+    PreparedQuery preparedQuery = datastore.prepare(query);
+    QueryResultList<Entity> userResult = preparedQuery.asQueryResultList(fetchOptions);
+
+    boolean isMaintainer = false;
+
+    if (userResult.size() == 1) {
```
mickrobk, line 56: Can you throw an error if the size() > 1?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -0,0 +1,81 @@
...
+      for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
+        isMaintainer = (boolean) entity.getProperty("isMaintainer");
+      }
+    }
+    GivrUser user = new GivrUser(userId, isMaintainer);
+    return user;
+  }
+
+  public static GivrUser getUserByEmail(String email) {
+    String authDomain = "gmail.com";
```
mickrobk, line 70: If you intend to **ever** support oauth, I'd suggest opening a tracking issue: "Support OAuth", and add a todo here that links to that issue

That way, you will be easily be able to find the places in the code where you assumed google auth.

mickrobk (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: I'll open an issue, but I'm not sure if we will be able to get it working within the next four weeks :)

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -0,0 +1,85 @@
...
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.PreparedQuery;
+
+public final class GivrUser {
+  private String id;
+  private boolean isMaintainer = false;
+
+  public GivrUser(String id, boolean isMaintainer) {
+    this.id = id;
+    this.isMaintainer = isMaintainer;
```
belraquib, line 37: I do think saving the email as an attribute is still helpful (just don't need to store it into the database); But i'd add that in another PR when needed.

belraquib (approved):

PR #33, opened 2020-07-08
add organization update functionality for forms

### LINKED ISSUE #35: Handle form submissions that have null value protection

Sarah-Addo:


src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +53,89 @@ public Organization(Entity entity) {
...
+        isApproved = true;
+      } else if(formValue.equals("notApproved")) {
+          isApproved = false;
+        }
+    } else if(propertyKey.equals("orgName")) {
+      name = formValue;
+    } else if(propertyKey.equals("orgEmail")) {
+      email = formValue;
+    } else if(propertyKey.equals("orgStreetAddress")) {
+      address = formValue;
```
andrewm898: Where are these variables initialized / stored? it looks like multiple functions reference them but I don't see them included as class variables

edit: nevermind, github just collapsed that code. Should we be using the this. keyword on these though?

Sarah-Addo: added this keyword

jnnybk (approved):

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +52,89 @@ public Organization(Entity entity) {
...
+      if(propertyKey.equals("isApproved") && !isMaintainer) {
+        continue;
+      }
+
+      // will only get moderatorList param if request is sent by a moderator or maintainer
+      if(propertyKey.equals("moderatorList") && !isModerator && !isMaintainer) {
+        continue;
+      }
+
+      // setting organization description is optional
```
belraquib: Why? Won't this clear out the organization description?

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +52,89 @@ public Organization(Entity entity) {
...
+        }
+      } else {
+        formValue = request.getParameter(formKey);
+      }
+      setOrganizationProperty(propertyKey, formValue);
+    }
+  }
+
+  public void commitOrganization(Entity entity) {
+    entity.setProperty("orgName", name);
```
belraquib: do you mean, this.name, this.email, etc?

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +52,89 @@ public Organization(Entity entity) {
...
+    entity.setProperty("orgDescription", description);
+    entity.setProperty("orgPhoneNum", phoneNum);
+    entity.setProperty("isApproved", isApproved);
+    entity.setProperty("orgUrl", urlLink);
+    entity.setProperty("moderatorList", moderators);
+  }
+
+  private String getParameterOrThrow(HttpServletRequest request, String formKey) {
+    String result = request.getParameter(formKey);
+    if(result.isEmpty()) {
```
belraquib: or null?

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +52,89 @@ public Organization(Entity entity) {
      this.urlLink = (String) entity.getProperty("orgUrl");
      this.moderators = (ArrayList) entity.getProperty("moderatorList");
   }
+
```
belraquib: instead of storing the data in separate variables as you do above (and the commit function where you set them directly), what about making these into functions?


example:
constructor() {
  this.entity = entity;
}

long id() {
   return (long) entity.getKey().getId();
}
String name() {
   return entity.getProperty("orgName");
}
etc

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +52,89 @@ public Organization(Entity entity) {
...
+  private String getParameterOrThrow(HttpServletRequest request, String formKey) {
+    String result = request.getParameter(formKey);
+    if(result.isEmpty()) {
+        throw new IllegalArgumentException("Form value cannot be null");
+    }
+    return result;
+  }
+
+  private void setOrganizationProperty(String propertyKey, String formValue) {
+    if(propertyKey.equals("moderatorList")) {
```
belraquib: If you store the entity, this can then instead be:

this.entity_.setProperty(propertyKey, formValue);

for everything except moderatorList and isApproved

ie;

if (propertyKey is moderatorList) {
  // what you have now
} else if (propertyKey is apprived) {
  // what you have now
} else {
  // this handles everything else.
  this.entity_.setProperty(propertyKey, formValue);
}

src/main/java/com/google/sps/data/Organization.java
belraquib: I think you don't need this function if you store just this.entity

src/main/java/com/google/sps/data/Organization.java
```diff
@@ -47,4 +52,89 @@ public Organization(Entity entity) {
...
+    properties.put("approval", "isApproved");
+    properties.put("moderator-list", "moderatorList");
+
+    for(Map.Entry<String, String> entry : properties.entrySet()) {
+      String formKey = entry.getKey();
+      String propertyKey = entry.getValue();
+      String formValue = "";
+
+      // will only get approval param is a maintainer sent the request
+      if(propertyKey.equals("isApproved") && !isMaintainer) {
```
belraquib: instead of checking for isApproved, what if you store a Set?

Set<String> requiresMaintainer; requiresMaintainer.add("a");

same for requestModerator


Then this code checks if the propertyKey is in the maintainer/moderator set and it's easy to add new fields that require permissions.

belraquib (changes requested):

src/main/java/com/google/sps/data/Organization.java
Sarah-Addo: added why we are doing this in the design doc

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public OrganizationUpdater(Entity entity) {
+     this.entity = entity;
+  }
+
+  public Entity getEntity() {
+      return this.entity;
+  }
+
+  public void updateOrganization(HttpServletRequest request, boolean isMaintainer, boolean isModerator) throws IllegalArgumentException{
```
mickrobk: What do you think of accepting a GiverUser instead of these booleans?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+      String formKey = entry.getKey();
+      String propertyKey = entry.getValue();
+      String formValue = "";
+
+      // will only get approval param is a maintainer sent the request
+      if(requiresMaintainer.contains(propertyKey) && !isMaintainer) {
+        continue;
+      }
+
+      // will only get moderatorList param if request is sent by a moderator or maintainer
```
mickrobk, line 75: "or" here doesn't line up with this if statement

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+    properties.put("org-phone-num", "orgPhoneNum");
+    properties.put("org-url", "orgUrl");
+    properties.put("org-description","orgDescription");
+    properties.put("approval", "isApproved");
+    properties.put("moderator-list", "moderatorList");
+
+    for(Map.Entry<String, String> entry : properties.entrySet()) {
+      String formKey = entry.getKey();
+      String propertyKey = entry.getValue();
+      String formValue = "";
```
belraquib: nit:  move to line 71

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+    if(result.isEmpty() || result == null) {
+        throw new IllegalArgumentException("Form value cannot be null");
+    }
+    return result;
+  }
+
+  private void setOrganizationProperty(String propertyKey, String formValue) {
+    if(propertyKey.equals("moderatorList")) {
+      ArrayList<String> newModeratorList = new ArrayList<String>(Arrays.asList(formValue.split("\\s*,\\s*")));
+      this.entity.setProperty("moderatorList", newModeratorList);
```
mickrobk, line 110: Is this right? Aren't you storing userids in the db, and accepting emails in the form?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+
+      // setting organization description is optional
+      if(!propertyKey.equals("orgDescription")) {
+        try {
+          formValue = getParameterOrThrow(request, formKey);
+        } catch(IllegalArgumentException err) {
+          throw new IllegalArgumentException();
+        }
+      } else {
+        formValue = request.getParameter(formKey);
```
belraquib: this can be null. what should your code do if it is null? set it to an empty string?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+  private void setOrganizationProperty(String propertyKey, String formValue) {
+    if(propertyKey.equals("moderatorList")) {
+      ArrayList<String> newModeratorList = new ArrayList<String>(Arrays.asList(formValue.split("\\s*,\\s*")));
+      this.entity.setProperty("moderatorList", newModeratorList);
+    } else if(propertyKey.equals("isApproved")) {
+      if(formValue.equals("approved")) {
+        this.entity.setProperty("isApproved", true);
+      } else if(formValue.equals("notApproved")) {
+          this.entity.setProperty("isApproved", false);
+        }
```
belraquib: check lint. spacing looks off

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private void setOrganizationProperty(String propertyKey, String formValue) {
+    if(propertyKey.equals("moderatorList")) {
+      ArrayList<String> newModeratorList = new ArrayList<String>(Arrays.asList(formValue.split("\\s*,\\s*")));
+      this.entity.setProperty("moderatorList", newModeratorList);
+    } else if(propertyKey.equals("isApproved")) {
+      if(formValue.equals("approved")) {
+        this.entity.setProperty("isApproved", true);
+      } else if(formValue.equals("notApproved")) {
```
mickrobk: do you need the if on the else here? Or can you say it is not approved if "approved" doesn't match?

mickrobk (changes requested):

belraquib (changes requested):
This looks really good! My only concern is how you'd handle null values for the org description

src/main/java/com/google/sps/data/OrganizationUpdater.java
belraquib: Good catch here.

src/main/java/com/google/sps/data/OrganizationUpdater.java
belraquib: an alternative is to assume the JS side sends true/false  -  I wonder if you can extend Andrew's  CoerceToBool change (moving into a more generic RequestHelper class) to accept what the expected "true" param would be.

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: So right now GiverUser does not include an IsModerator check or field so I would have to add stuff to GiverUser which is still a work in progress in order to test. Should I go ahead and do it anyways or add a TODO?

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: I think it does because it would do the continue if a request was not sent by a moderator or a maintainer. The check would fail if it was a moderator or then fail if it was a maintainer. I think the key here is that the isModerator check is before the isMaintainer check.

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: Now will be using givr user translate email to ids function

src/main/java/com/google/sps/data/OrganizationUpdater.java
mickrobk: OK, I think you're right. Is there a way to rewrite this so there are less "nots"?
Eg currently this is
allow access = not(a && not(b) && not(c))

src/main/java/com/google/sps/data/OrganizationUpdater.java
mickrobk: There is currently a GiverUser checked in right; what are your concerns here? Is there a middle ground option?

mickrobk: To be clear: the heart of my question is basically: if we have this abstraction for "User," why aren't we using it and extending it? I want to make sure the work we do complements each other; that we're not just pulling in different directions.

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: yes i think its currently a && !b && !c so it would be  a && not( b || c) by demorgans law

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: My concern is that at the moment there GivrUser does not have a way of fulfilling the isModerator boolean check. On the edit org servlet side it would pass in request, GivrUser.isMaintainer, true since the edit functionality would only be displayed to users who are at least moderators in the first place. But when it comes from the Add Organization side it does not care who the user is since the registration form only takes care of certain params which is why it calls it with request, false, false. If I replaced this with a GiverUser then isMaintainer and isModerator would both have to result in false in order for the isApproved and ModeratorList params to not be checked which could be false since they could be a maintainer and they would technically be the first moderator for the org that they are submitting the request for. The way im looking at it now these checks are more for specifying which servlet is calling this function. I am thinking that if we did pass in a GiverUser param then we would also have to pass in if its calling update for registering or not so it ignores the user which seems like what makes the most sense

mickrobk: It seems like an error prone pattern to ask the caller of this function to sometimes lie about the authorization level of the user! Is there another way we could solve this?

belraquib: Consider: right now you are planning on sending down just "true" for isModerator right?

If we assume GIvrUser will do this can you can store this as local variables. For now you can always set isModerator to false.

On the consideration of different behavior for the caller: you should be careful about using  isModerator/isMaintainer to signify who the caller is. If you really have to change behavior based on who is calling the function, then consider a different parameter. If you did want to continue to use the call as if the caller is always unauthenticated, then you can create a subclass of the GivrUser or somehow consruct a GivrUser that guarantees it has no permissions.

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,127 @@
...
+    properties.put("org-url", "orgUrl");
+    properties.put("org-description","orgDescription");
+    properties.put("approval", "isApproved");
+    properties.put("moderator-list", "moderatorList");
+
+    for(Map.Entry<String, String> entry : properties.entrySet()) {
+      String propertyKey = entry.getValue();
+
+      // if updating for registering an organization then do not want to consider fields that require maintainer or moderator permissions
+      if(forRegistration && (requiresModerator.contains(propertyKey) || requiresMaintainer.contains(propertyKey))) {
```
belraquib: this will be a bit simpler to read if you used some helper vars;


bool propertyRequiresMaintainer = requiresMaintainer.contains(propertyKey);
bool propertyRequiresModerator = requiresModerator.contains(propertyKey);

Then the rest of the code will read a bit easier (and you won't recompute the set membership)

belraquib: I would even add an additional:

bool propertyRequiresAuth = propertyRequiresMaintainer || propertyRequiresModerator;

now this condition becomes

if (forRegistration && propertyRequiresAuth) {
  continue;
}

which really makes it easy to read

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -0,0 +1,127 @@
...
+      }
+    } else {
+        this.entity.setProperty(propertyKey, formValue);
+    }
+  }
+
+  private ArrayList<String> translateEmailsToIds(ArrayList<String> emails) {
+    ArrayList<String> userIds = new ArrayList<String>();
+    for(String email : emails) {
+      GivrUser newUser = GivrUser.getUserByEmail(email);
```
belraquib, line 127: can the user returned ever be null? or can the email address fail to generate a user id (from a typo, for instance!).

belraquib (changes requested):

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: reading the documentation it seems that it will always make a new user even if a typo exists so that unique id will be added to the list

belraquib: Looks like it can return null

https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/users/User.html#getuserid

Reading that, I'm not sure if we can get the user from an email address. Can you file an issue to investigate?

belraquib (approved):

PR #38, opened 2020-07-08
Add function to GivrUser.java to add new logged in user into Datastore

jnnybk:
1) Rename AddUserServlet to AuthenticateServlet
2) Add functionality to GivrUser and remove repetitive code


src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -57,11 +66,13 @@ public static GivrUser getUserByIdFromDatastore(String userId) {
 
     boolean isMaintainer = false;
 
-    if (userResult.size() == 1) {
+    if (userResult.size() < 1) {
+      addNewUserToDatastore(userId, isMaintainer);
```
mickrobk: Why do we need to do this?

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -37,61 +37,33 @@
...
-        for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
-          isUserMaintainer = (boolean) entity.getProperty("isMaintainer");
-        }
-        currentUser = new GivrUser(userId, isUserMaintainer);
-      }
+    GivrUser user = GivrUser.getLoggedInUser();
+    boolean isLoggedIn;
+    String url;
+    // User is null when they are not logged in.
+    if (user == null) {
```
mickrobk: Instead of null here, could you simplify by merging GiverUser with UserLoginInfo?

mickrobk (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: This stores each new User to the Datastore. I chose to implement this here in GivrUser.java instead of in the AuthenticateServlet.

mickrobk: Why do you need to store each user in the datastore?

belraquib: I have a different question: would you expect a  function that starts with Get* mutates the datastore?

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
jnnybk: Would that mean removing the isLoggedIn boolean variable in UserLoginInfo.java, adding the url string variable to GivrUser.java, and removing UserLoginInfo.java completely?

mickrobk: What if you moved both the isloggedin and loginurl to giveruser? Then you'd also be able to represent a non-logged in user without using null

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: To Rob, I decided to still store users as User entities in the Datastore to track their Maintainer status, instead of creating the Maintainer entity that we also discussed in the pod meeting. The podmates and I discussed that we may need to use this implementation and add onto this in the future.

To Tay, I fixed it to doPost. That makes more sense.

belraquib: What I meant was this function is called "Get"UserById. That word get usually implies that a function should be constant relative to any backing data. It's odd that the user in the datastore is created here.

belraquib (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
belraquib: In terms of tracking maintainer status:

If a user isn't in the datastore does that imply that the user is a maintainer or does it mean that the user is not a maintainer? Can you use the non-presence as a signal?

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -0,0 +1,53 @@
...
+
+@WebServlet("/authenticate")
+public class AuthenticateServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse  response) throws IOException {
+    response.setContentType("application/json;");
+    Gson gson = new Gson();
+
+    UserService userService = UserServiceFactory.getUserService();
```
belraquib: you don't need this anymore.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -65,7 +71,8 @@ public static GivrUser getUserByIdFromDatastore(String userId) {
       throw new IllegalArgumentException("More than one user with the userId was found.");
     }
 
-    GivrUser user = new GivrUser(userId, isMaintainer);
+
+    GivrUser user = new GivrUser(userId, isMaintainer, isLoggedIn, "" /* URL will be set in AuthenticateServlet */);
```
belraquib: maybe reword the comment to mention you don't need the url if you are already logged in.

belraquib (approved):

PR #39, opened 2020-07-09
add isModerator function to GivrUser

### LINKED ISSUE #35: Handle form submissions that have null value protection

Sarah-Addo:


src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -46,6 +46,11 @@ public boolean isMaintainer() {
     return this.isMaintainer;
   }
 
+  //TODO(): get correct moderator status for organization
+  public boolean isModerator(String organization) {
```
jnnybk: Would organization represent organization's ID?

Sarah-Addo: yes, will modify to say organizationID

belraquib (approved):

PR #42, opened 2020-07-10
Set up filtering frontend

andrewm898:
This PR adds frontend functionality to the "enter zipcode" section and replaces the "filter" button with a datalist element
Here is the current appearance of this frontend: 
image

****This functionality sends the parameters to the backend correctly formatted, but the backend does not modify the query with those parameters yet***

Current user flow/abilities: 
-The user can enter a zipcode and hit enter, this adds that zipcode to the urlsearchparams and calls getListOfOrganizations with that zipcode added. If the user wants to enter another zipcode, the previous one they added will still be in that text field, so they will need to backspace it and hit enter again. This new 5 digit sequence will replace the old zipcode in the urlsearchparams.

-The user can either type their own, or select one of the suggested filtering keywords from the datalist dropdown. After the user is satisfied with the word in the datalist textfield, they must hit enter to send the request. **Important difference from the zipcode functionality:** : After hitting enter on a filter keyword in the datastore, the user CAN enter another keyword and the previous filtering keyword will NOT be overwritten. They will be sent as an array, and to see how this array is referenced on the backend look at line 57 of ListOrg servlet. The urlsearchparams can contain a single zipcode and any amount of strings with "filterParam" keys 

I intentionally left the print statements in the list org servlet so there is no confusion on how to access these url parameters.

Remaining TODOs for frontend filtering:
-Display all current active filters being applied to the query for the user to see on the frontend
-Allow the user to selectively remove filters from the urlsearchparams after adding them
-Add debouncing protection to the enter buttons / text field

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -7,3 +7,4 @@ indexes:
     direction: desc
   - name: isApproved
   - name: moderatorList
+  
--- a/index.yaml
+++ b/index.yaml
@@ -1,0 +1,9 @@
+indexes:
+
+- kind: Distributor
+  ancestor: no
+  properties:
+  - name: creationTimeStampMillis
+    direction: desc
+  - name: isApproved
+  - name: moderatorList
--- a/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
@@ -48,18 +48,6 @@ public class ListOrganizationsServlet extends HttpServlet {
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     GivrUser currentUser = GivrUser.getLoggedInUser();
 
-    /* These statements show how to reference the parameters being received, for easier implementation of this backend */
-    if (request.getParameter("zipcode") != null) {
-      System.out.println("Zipcode is: " + request.getParameter("zipcode"));
-    }
-
-    if (request.getParameterValues("filterParam") != null) {
-      String filterParams[] = request.getParameterValues("filterParam");
-      for (String param : filterParams) {
-        System.out.println("One of the filter params is: " + param);
-      }
-    }
-
     /* All get requests will return a maximum of 5 organization entities */
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(5);
     Query query = getQueryFromParams(request, currentUser);
--- a/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
@@ -36,7 +36,6 @@ import com.google.sps.data.Organization;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.sps.data.GivrUser;
-import java.lang.*;
 import java.io.IOException;
 
 @WebServlet("/list-organizations")
--- a/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
@@ -36,6 +36,7 @@ import com.google.sps.data.Organization;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.sps.data.GivrUser;
+import java.lang.*;
 import java.io.IOException;
 
 @WebServlet("/list-organizations")
@@ -48,9 +49,20 @@ public class ListOrganizationsServlet extends HttpServlet {
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     GivrUser currentUser = GivrUser.getLoggedInUser();
 
+    /* These statements show how to reference the parameters being received, for easier implementation of this backend */
+    if (request.getParameter("zipcode") != null) {
+      System.out.println("Zipcode is: " + request.getParameter("zipcode"));
+    }
+
+    if (request.getParameterValues("filterParam") != null) {
+      String filterParams[] = request.getParameterValues("filterParam");
+      for (String param : filterParams) {
+        System.out.println("One of the filter params is: " + param);
+      }
+    }
+
     /* All get requests will return a maximum of 5 organization entities */
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(5);
-
     Query query = getQueryFromParams(request, currentUser);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     PreparedQuery prepQuery = datastore.prepare(query);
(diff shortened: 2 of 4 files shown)


index.yaml
```diff
@@ -0,0 +1,9 @@
+indexes:
+
+- kind: Distributor
+  ancestor: no
+  properties:
+  - name: creationTimeStampMillis
+    direction: desc
+  - name: isApproved
+  - name: moderatorList
```
Sarah-Addo: new line

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -36,6 +36,7 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.sps.data.GivrUser;
+import java.lang.*;
```
Sarah-Addo: this imports everything, why did you want to do this

src/main/webapp/search_area.js
```diff
@@ -40,53 +44,75 @@ class SearchArea {
...
+    for (let i = 0; i < this.filterOptions.length; i++) {
+      const option = document.createElement("option");
+      option.value = this.filterOptions[i];
+      this.filterDataList.appendChild(option);
+    }
+
+    this.filterInputArea.appendChild(this.filterDataList);
+    this.organizationSearchArea.appendChild(this.filterInputArea);
+
+    this.organizationListDiv = document.createElement("div");
```
Sarah-Addo: rename to organizationListArea to be inline with the other naming conventions>

src/main/webapp/search_area.js
```diff
@@ -103,11 +129,24 @@ class SearchArea {
...
+      this.organizationListDiv.appendChild(newOrganization.getOrganization());
     });
+  }
 
-    this.organizationSearchArea.appendChild(this.organizationList);
-    this.searchArea.appendChild(this.organizationSearchArea);
-    this.searchArea.appendChild(this.organizationPopupArea);
+  async handleZipcodeSubmission(zipcodeForm) {
+    this.filterParams.set("zipcode", zipcodeForm.zipcode.value);
+    this.organizationsObjectsList = [];
```
Sarah-Addo: nit: organizationObjectsList?

src/main/webapp/search_area.js
```diff
@@ -103,11 +129,24 @@ class SearchArea {
...
+    this.organizationListDiv.innerHTML = "";
```
Sarah-Addo: use .textContent it is more secure than .innerHTML

andrewm898: I used innerHTML here bc the last time I implemented code that clears existing divs like this I was told to change it to innerHTML instead of textContent

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: My bad- forgot to take this out after trying to figure something out.

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -87,11 +97,10 @@ public Query getQueryFromParams(HttpServletRequest request, GivrUser currentUser
       query.setFilter(new FilterPredicate("moderatorList", FilterOperator.EQUAL, userId));
     }
```
andrewm898, line 99: The only change here was uncommenting out this code now that indexing is working*

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -48,9 +48,20 @@
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     GivrUser currentUser = GivrUser.getLoggedInUser();
 
+    /* These statements show how to reference the parameters being received, for easier implementation of this backend */
+    if (request.getParameter("zipcode") != null) {
```
belraquib: It's good practice to remove these types of statements when submitting unless you really need it.

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -72,7 +82,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
   /* This function constructs a query based on the request parameters & user's role */
   public Query getQueryFromParams(HttpServletRequest request, GivrUser currentUser) {
-    Query query = new Query("Distributor").addSort("creationTimeStampMillis", SortDirection.DESCENDING);;
+    Query query = new Query("Distributor").addSort("creationTimeStampMillis", SortDirection.DESCENDING);
```
belraquib, line 73: nit: this and below could have been in another PR.

src/main/webapp/search_area.js
```diff
@@ -12,81 +12,96 @@
...
+
+    this.filterInputArea = document.createElement("input");
+    this.filterInputArea.setAttribute("list", "filter-datalist");
+    this.filterInputArea.setAttribute("id", "filter-input-area");
+    this.filterInputArea.setAttribute("placeholder", "Filter Results");
+    this.filterInputArea.addEventListener('keypress', function (e) {
+      if (e.key === 'Enter') {
+        this.setUrlParamValue("filterParam", this.filterInputArea.value);
+      }
+    }.bind(this));
```
belraquib: please use a lambda instead of bind

(e) => {

}

src/main/webapp/search_area.js
```diff
@@ -12,81 +12,96 @@
...
+
+    this.filterDataList = document.createElement("datalist");
+    this.filterDataList.setAttribute("id", "filter-datalist");
+    this.filterOptions = ["Foods", "Clothing", "Shelter"];
+    for (let i = 0; i < this.filterOptions.length; i++) {
```
belraquib: If you only refer to values but never the index you can write

for (const value of this.filterOptions) {
  const option = ..;
  option.value = value;
  ...
}

src/main/webapp/search_area.js
```diff
@@ -12,81 +12,96 @@
...
     this.organizationPopupArea.classList.add("hide-popup");
 
-    organizations.forEach((organization) => {
-      const newOrganization = new Organization(organization, isMaintainer);
+    this.organizationSearchArea.appendChild(this.organizationListArea);
+    this.searchArea.appendChild(this.organizationSearchArea);
+    this.searchArea.appendChild(this.organizationPopupArea);
+  }
+
+  async renderListOfOrganizations() {
```
belraquib: why is this an async function?

belraquib (changes requested):

src/main/webapp/search_area.js
andrewm898: I just realized this was unnecessary

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: next time I'll be more careful about separating these

belraquib (approved):

PR #44, opened 2020-07-10
Refactor add-maintainer servlet, implement authenticate servlet, modify GivrUser class, and re-structure Datastore entity "User"

### LINKED ISSUE #41: Figure out a way to get a user id with an invited email
Right now we depended on getting a user id from an email but that would only work if the user has previously logged in

jnnybk:
Following the pod's discussion, I implemented TODOs that identifies where code needs to be modified. The notes from the discussion and our solution plan can be found here: https://docs.google.com/document/d/1sN3mAo623Se6gs7sWC13P4w9XBJ9U2smvHpefcMUykQ/edit?usp=sharing. 

(EDIT)
This PR completes the following features:
1) Modify GivrUser class
2) Implement the flow of AuthenticateServlet 
3) Implement AddMaintainerServlet


src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.PreparedQuery;
+
+@WebServlet("/add-maintainer")
+public class AddMaintainerServlet extends HttpServlet {
+
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/json;");
+
+    GivrUser currUser = GivrUser.getLoggedInUser();
```
jnnybk: should be getCurrentLoggedInUser(). Not complete yet.

andrewm898:
Could you resolve the merge conflicts in GivrUser.java?

src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
```diff
@@ -0,0 +1,84 @@
...
+
+    // User should be able to add another Maintainer only if they are a Maintainer.
+    if (!currUser.isMaintainer()) {
+      String newMaintainerEmail = request.getParameter("userEmail");
+
+      changeMaintainerStatus(newMaintainerEmail);
```
Sarah-Addo: you are changing the user status if the current user is not a maintainer, did you mean to do the opposite

src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
```diff
@@ -0,0 +1,84 @@
...
+    Query query = new Query("User").setFilter(queryFilter);
+      
+    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(1);
+    PreparedQuery preparedQuery = datastore.prepare(query);
+    QueryResultList<Entity> userResult = preparedQuery.asQueryResultList(fetchOptions);
+
+    boolean isMaintainer = false;
+    String userId = "";
+
+    // If User with email does not exist in the User table, this means that the User is not a Maintainer.
```
Sarah-Addo: update comment

src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
jnnybk: yes! it was for testing purposes :) it's been changed back!

src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
jnnybk: updated!

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,48 +53,118 @@ public boolean isMaintainer() {
...
-  public static GivrUser getUserByIdFromDatastore(String userId) {
+  public String getUserEmail() {
+    return this.email;
+  }
+
+  public boolean isLoggedIn() {
+    return this.isLoggedIn;
+  }
+
+  public static PreparedQuery getPreparedQueryResultUserWithProperty(String propertyName, String propertyValue) {
```
belraquib: Can you leave a comment about how this is meant to be used?  I'd imagine you want to be able to have multiple query filters in a prepared query, ie;

PrepartedQuery = GiverUser.PreparedQueryBuilder().With("userId", "fjosfjs").With("isMaintainer", true).Build();

?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,48 +53,118 @@ public boolean isMaintainer() {
...
     Query query = new Query("User").setFilter(queryFilter);
-      
-    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(1);
+
     PreparedQuery preparedQuery = datastore.prepare(query);
+    return preparedQuery;
+  }
+
+  // Check if User with propertyName, propertyValue exists within Datastore.
+  public static boolean checkIfUserWithPropertyExists(String propertyName, String propertyValue) {
```
belraquib: I would simplify this to ;

private static boolean ensureSingleResult(QueryResultList<Entity> userResult).

That way you only make the request once and use the same result list for getting the user as well as this check. Any reason to make the function public ?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,48 +53,118 @@ public boolean isMaintainer() {
...
+    if (userResult.size() < 1) {
+      return false;
+    } else if (userResult.size() > 1) {
+      throw new IllegalArgumentException("More than one user with the " + propertyName + " was found.");
+    }
+    return true;
+  }
+
+  // Updates a User entity in Datastore, identifying with the first two parameters with values from second two parameters.
+  public static void updateUserInDatastore(String identifyingProperty, String identifyingValue, String updateProperty, String updateValue) {
```
belraquib: What other parameters would you expect for identifyingProperty? Why limit to a single update property?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,48 +53,118 @@ public boolean isMaintainer() {
...
+    for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
+      Transaction txn = datastore.beginTransaction();
+      try {
+        entity.setProperty(updateProperty, updateValue);
+        datastore.put(txn, entity);
+
+        txn.commit();
+      } finally {
+        if (txn.isActive()) {
+          txn.rollback();
```
belraquib: should you throw an exception here to tell the caller that the update failed?

src/main/webapp/maintainer.js
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+document.addEventListener('DOMContentLoaded', () => {
+  // TODO: Add UI for adding Maintainer as a Maintainer. Create input box to submit on click and call /add-maintainer.
+});
+
+async function getMaintainerStatus(email) {
```
belraquib: this isn't really getting maintainer status, but adding the maintainer as a user? can you make this a POST request?

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -44,9 +44,31 @@ public void doGet(HttpServletRequest request, HttpServletResponse  response) thr
...
+        if (!isMaintainer) { // User is a Moderator.
+
+          // TODO: Update organization.
+        }
+
+        // Update entity with current user's userId.
+        GivrUser.updateUserInDatastore("userEmail", user.getUserEmail(), "userId", user.getUserId());
+      }
+    } else {
+      // Ensures that the userEmail is up-to-date.
```
belraquib: should you check if the userEmail != the stored email? Otherwise you'll have extra writes and writes are more expensive than reads.

belraquib (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: Comment added!

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: This function currently is used in AddMaintainerServlet when checking if the new Maintainer exists within the Datastore. It's also used in the AuthenticateServlet to check if the Id exists; if not, then it checks if the email exists.

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: Does this mean to have a method that can take a variable number of properties?

So, something like updateUserInDatastore(String identifyingProperty, String identifyingValue, Map<String, String> propertiesAndValuesToUpdate)?

belraquib: Yup.

Any reason to make these statics?

Why not:

GivrUser user = GivrUser.GetUserFromProperty("userId", foobar);
user.setEmail("email");
user.commitInTransaction(); 

?

jnnybk: I made this function static because the entity that's being changed would depend on the identifying property, not the object that this function is called on (if this function is not static).

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +55,119 @@ public boolean isMaintainer() {
...
-  public static GivrUser getUserByIdFromDatastore(String userId) {
+  /*
+   * This function returns the PreparedQuery for the "User" entities that have the propertyName, propertyValue.
+   * Use case: PreparedQuery preparedQuery = GivrUser.getPreparedQueryResultOfUserWithProperty("userId", "abcdefg");
+   */
+  public static PreparedQuery getPreparedQueryResultOfUserWithProperty(String propertyName, String propertyValue) {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Filter queryFilter = new FilterPredicate("userId", FilterOperator.EQUAL, userId);
+    Filter queryFilter = null;
+    if (propertyName.equals("isMaintainer")) {
```
belraquib: When would you call this function with an isMaintainer property?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +55,119 @@ public boolean isMaintainer() {
...
+
     PreparedQuery preparedQuery = datastore.prepare(query);
-    QueryResultList<Entity> userResult = preparedQuery.asQueryResultList(fetchOptions);
+    return preparedQuery;
+  }
 
-    boolean isMaintainer = false;
-    boolean isLoggedIn = true;
+  // Check if User with propertyName, propertyValue exists within Datastore.
+  public static boolean checkIfUserWithPropertyExists(String propertyName, String propertyValue) {
```
belraquib: Entity getUserFromDatabaseWithProperty(String propertyName, String propertyValue) {
  // either return null or throw an exception if it doesn't exist.
}

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +55,119 @@ public boolean isMaintainer() {
...
+    boolean isMaintainer = false;
+    if (entity != null) {
+      userId = (String) entity.getProperty("userId");
+      isMaintainer = (boolean) entity.getProperty("isMaintainer");
+    }
+    return new GivrUser(userId, isMaintainer, false /* isLoggedIn */, "" /* loginURL */, email);
   }
 
-  public static GivrUser getLoggedInUser() {
+  public static GivrUser getCurrentLoggedInUser() {
```
belraquib, line 140: note that the email returned in the givruser is what's in the database, NOT what's in the current user!

src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
```diff
@@ -0,0 +1,83 @@
...
+@WebServlet("/add-maintainer")
+public class AddMaintainerServlet extends HttpServlet {
+
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/json;");
+
+    GivrUser currUser = GivrUser.getCurrentLoggedInUser();
+
+    // User can add another Maintainer only if they are a Maintainer.
+    if (currUser.isMaintainer()) {
```
belraquib: if (!currUser.isMaintainer) {
  return NOT FOUND
}

// All my complicated is maintainer code...

src/main/java/com/google/sps/servlets/AddMaintainerServlet.java
```diff
@@ -0,0 +1,83 @@
...
+    newUserEntity.setProperty("userId", "");
+    newUserEntity.setProperty("isMaintainer", true);
+    newUserEntity.setProperty("userEmail", email);
+
+    datastore.put(newUserEntity);
+  }
+
+  public void changeMaintainerStatus(String email, String userId) {
+    boolean isMaintainer = true;
+    Map<String, String> propertyNamesAndValuesToUpdate = new HashMap<String, String>();
```
belraquib: could you store Map<String, Object> ?

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -44,8 +46,38 @@ public void doGet(HttpServletRequest request, HttpServletResponse  response) thr
...
+
+    boolean doesUserIdExistInDatastore = GivrUser.checkIfUserWithPropertyExists("userId", user.getUserId());
+    if (!doesUserIdExistInDatastore) {
+      boolean doesUserEmailExistInDatastore = GivrUser.checkIfUserWithPropertyExists("userEmail", user.getUserEmail());
+      if (doesUserEmailExistInDatastore) {
+        // User is either a Moderator or Maintainer, that has not logged in.
+        boolean isMaintainer = user.isMaintainer();
+        if (!isMaintainer) { // User is a Moderator.
+
+          // TODO: Update organization.
```
belraquib, line 54: I expected a return here since you don't want to add an entry if they are not a maintainer, right?

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -44,8 +46,38 @@ public void doGet(HttpServletRequest request, HttpServletResponse  response) thr
...
 
-    GivrUser user = GivrUser.getLoggedInUser();
+    GivrUser user = GivrUser.getCurrentLoggedInUser();
 
+    Map<String, String> propertyNamesAndValuesToUpdate = new HashMap<String, String>();
+
+    boolean doesUserIdExistInDatastore = GivrUser.checkIfUserWithPropertyExists("userId", user.getUserId());
+    if (!doesUserIdExistInDatastore) {
+      boolean doesUserEmailExistInDatastore = GivrUser.checkIfUserWithPropertyExists("userEmail", user.getUserEmail());
+      if (doesUserEmailExistInDatastore) {
```
belraquib: if (!userInDatastore) {
   MaybeUpdateUserByEmailInDatastore(user.email);
} else {
   MaybeUpdateEmailAddressOfUserIdInDatastore(user);  // passing in the givruser
}

belraquib (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: I chose to return null here!

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
jnnybk: We want to add an entry if the User is a Maintainer or a Moderator, and if we return here, then we wouldn't be able to add Moderators into our Datastore!

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +56,104 @@ public boolean isMaintainer() {
...
-  public static GivrUser getUserByIdFromDatastore(String userId) {
+  public static boolean checkIfUserWithPropertyExists(String propertyName, String propertyValue) {
+    if (getUserFromDatastoreWithProperty(propertyName, propertyValue) == null) {
+      return false;
+    }
+    return true;
+  }
+
+  // Gets User with propertyName, propertyValue exists within Datastore.
+  public static Entity getUserFromDatastoreWithProperty(String propertyName, String propertyValue) {
```
jnnybk, line 76: I left this as a public function because it is used in AuthenticateServlet in order to update the User entity.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +56,104 @@ public boolean isMaintainer() {
...
+    return this.isLoggedIn;
+  }
+
+  // TODO(): get correct moderator status for organization
   public boolean isModerator(long organizationId) {
       return false;
   }
 
-  public static GivrUser getUserByIdFromDatastore(String userId) {
+  public static boolean checkIfUserWithPropertyExists(String propertyName, String propertyValue) {
```
belraquib: recommend removing this fiunction and just using

getUserFromDataStore(...) == null

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +56,97 @@ public boolean isMaintainer() {
...
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Filter queryFilter = new FilterPredicate("userId", FilterOperator.EQUAL, userId);
+
+    Filter queryFilter = new FilterPredicate(propertyName, FilterOperator.EQUAL, propertyValue);
     Query query = new Query("User").setFilter(queryFilter);
-      
-    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(1);
     PreparedQuery preparedQuery = datastore.prepare(query);
-    QueryResultList<Entity> userResult = preparedQuery.asQueryResultList(fetchOptions);
+    Entity entity = preparedQuery.asSingleEntity();
```
belraquib: IIRC this throws if there is more than one result. You can decide if you want to handle that here or not

https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +56,97 @@ public boolean isMaintainer() {
...
+    Entity entity = getUserFromDatastoreWithProperty(identifyingProperty, identifyingValue);
+
+    if (entity == null) {
+      throw new Error("User with " + identifyingProperty + ": " + identifyingValue + " was not found.");
+    }
 
-    if (userResult.size() == 1) {
-      for (Entity entity: preparedQuery.asIterable(fetchOptions)) {
-        isMaintainer = (boolean) entity.getProperty("isMaintainer");
+    Transaction txn = datastore.beginTransaction();
```
belraquib: Any reason why you are doing a transaction here? You aren't updating multiple entities? (IIRC, the write doesn't happen until you call put())

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -76,8 +76,13 @@ public static Entity getUserFromDatastoreWithProperty(String propertyName, Strin
     Filter queryFilter = new FilterPredicate(propertyName, FilterOperator.EQUAL, propertyValue);
     Query query = new Query("User").setFilter(queryFilter);
     PreparedQuery preparedQuery = datastore.prepare(query);
-    Entity entity = preparedQuery.asSingleEntity();
 
+    Entity entity = null;
+    try {
+      entity = preparedQuery.asSingleEntity();
+    } catch(PreparedQuery.TooManyResultsException exception) {
+      exception.printStackTrace();
```
belraquib: Instead of printing a stack trace, can you log an error message?

src/main/java/com/google/sps/data/GivrUser.java
belraquib: Can you answer this question?

belraquib (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: I think originally I had used transactions for just proof of concept but it makes sense not to use them here :) will remove!

Sarah-Addo (approved):

belraquib (approved):

PR #45, opened 2020-07-10
Filtering backend

andrewm898:
This enables filtering via zip code by adding a zip code field to all organization-related classes / servlets where appropriate, and setting up composite filtering to enable this filter to be applied along with the existing filters in the servlet. The code is currently deployed and works to filter by zip code

note: The ability to remove all urlSearchParams and the handling of null entries in the zipcode form are still TODO's, so if you try it out, you can enter different zip codes to see different results as normal, but if you want to go back to seeing all organizations the only way to do that right now is to refresh the page


src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -72,25 +74,47 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   public Query getQueryFromParams(HttpServletRequest request, GivrUser currentUser) {
     Query query = new Query("Distributor").addSort("creationTimeStampMillis", SortDirection.DESCENDING);
 
+    ArrayList<Filter> filterCollection = new ArrayList<Filter>();
+
+    String zipcode  = "";
+    boolean queryForZipcode = false;
+    if (request.getParameter("zipcode") != null) {
+      zipcode = request.getParameter("zipcode");
+      queryForZipcode = true;
```
belraquib, line 83: Do you want to verify this zipcode? (It's OK to say no, as the search will fail, but consider what will help the user understand and fix their mistake)

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -72,25 +74,47 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
     if (isUserLoggedIn && displayUserOrgs) {
       /* If the user is logged in and wants to just see their orgs, get their user ID & index with it*/
       String userId = currentUser.getUserId();
-      query.setFilter(new FilterPredicate("moderatorList", FilterOperator.EQUAL, userId));
+      filterCollection.add(new FilterPredicate("moderatorList", FilterOperator.EQUAL, userId));
     }
 
-    if (!userIsMaintainer) {
+    if (userIsMaintainer) {
```
belraquib: The comment below and this check don't agree - it states that you want non-maintainers to only see approved orgs.

belraquib: This is getting complicated enough that you should consider a test:

See:

https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting#datastore-memcache

belraquib (changes requested):

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: You're right, this should be (!userIsMaintainer), I changed it when trying to debug something and forgot to change it back. I agree that unit testing would definitely be useful for this, and I'm going to implement it for this servlet

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: I was thinking we could display "No Results Found" if nothing comes back for the query, which would apply if they enter a bad zip code. I added a point to the filtering frontend issue / todo's list to address this

belraquib (approved):

PR #49, opened 2020-07-13
Refactor organization updater to update non form properties

Sarah-Addo:


src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -57,6 +60,7 @@ public void updateOrganization(HttpServletRequest request, GivrUser user, boolea
     properties.put("approval", "isApproved");
     properties.put("moderator-list", "moderatorList");
 
+    // Updates entity properties from form
```
belraquib, line 62: What if instead of assuming the history update in the helper, you took in an EmbeddedEntity instance (for the history) here?

belraquib (changes requested):

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -129,4 +137,35 @@ private void setOrganizationProperty(String propertyKey, String formValue) {
     }
     return userIds;
   }
+
+  private void updateNonFormProperties(GivrUser user, boolean forRegistration, EmbeddedEntity historyUpdate) {
+    /* MillisecondSinceEpoch represent the number of milliseconds that have passed since
+     * 00:00:00 UTC on January 1, 1970. It ensures that all users are entering a representation
+     * of time that is independent of their time zone */
+    long millisecondSinceEpoch = Instant.now().toEpochMilli();
```
belraquib: can you get the time for the update from the historyUpdate instead of now?

belraquib (approved):

PR #50, opened 2020-07-13
Unit tests for filtering

andrewm898:
Adds a ListOrg unit test file & necessary dependencies to accomplish these unit tests / libraries used

NOTE: **The build failure in this is because I separated this PR from the PR that adds all of the new code for the zipcode test in ListOrganizationServlet, so that test case is failing here. As long as this file is merged AFTER #45, this build error will not be an issue.


src/test/java/com/google/sps/ListOrgServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+ 
+  /* The masterEntityList is a list of entities that are populated into the local datastore, and also exist 
+   * in an arraylist. For each test, the datastore applies the Query received from the servlet, and then
+   * compares that query to a hardcoded list of expected entities given that tests parameters / user fields */
+  private ArrayList<Entity> masterEntityList = new ArrayList<>();
+ 
+  @Before
+  public void setUp() {
+    helper.setUp();
+
```
belraquib: A top level comment as an ascii table would be helpful to understand the test below.

src/test/java/com/google/sps/ListOrgServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+ 
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+ 
+  @Test
+  public void testQueryWithNoFilter() {
+    /* For this test, all of the entities need to be returned because there are no filter limits,
+     * but the master list must be reversed */
```
belraquib: why is the master list reversed?

belraquib: [ie, be explicit about your expectation of the sort order]

src/test/java/com/google/sps/ListOrgServletTest.java
```diff
@@ -0,0 +1,179 @@
...
+ 
+    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(10);
+    Assert.assertArrayEquals(expectedList.toArray(), datastore.prepare(receivedQuery).asList(fetchOptions).toArray());
+  }
+  @Test
+  public void testQueryAsNormalUser() {
+ 
+    ArrayList<Entity> expectedList = new ArrayList<Entity>();
+    /* Only the isApproved = true orgs are added to this simulated query, and they are added in
+     * descending order of their timestamp */
```
belraquib, line 156: good comment :)

belraquib (changes requested):

src/test/java/com/google/sps/ListOrgServletTest.java
```diff
@@ -0,0 +1,197 @@
...
+ 
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+ 
+  @Test
+  public void testQueryWithNoFilter() {
+    /* For this test, all of the entities need to be returned because there are no filter limits,
+     * but the master list must be reversed since the query returns them in descending order*/
```
belraquib: I think you can simplify this comment to:

Because there are no filters, all entities are returned in reverse time order.

belraquib (approved):

PR #52, opened 2020-07-14
Refactor GivrUser.java

jnnybk:
Build will fail because this PR commits only the GivrUser.java file. This PR also modifies servlets that use the GivrUser method getCurrentLoggedInUser() from getLoggedInUser().


src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -51,53 +56,104 @@ public boolean isMaintainer() {
...
-  public static GivrUser getUserByIdFromDatastore(String userId) {
+  public static boolean checkIfUserWithPropertyExists(String propertyName, String propertyValue) {
+    if (getUserFromDatastoreWithProperty(propertyName, propertyValue) == null) {
+      return false;
+    }
+    return true;
+  }
+
+  // Gets User with propertyName, propertyValue exists within Datastore.
+  private static Entity getUserFromDatastoreWithProperty(String propertyName, String propertyValue) {
```
jnnybk, line 80: I'm changing this back to a public function because it's needed in AuthenticateServlet when updating the User entity.

Issue #46, opened 2020-07-10
Filtering Frontend improvements


andrewm898:
The filtering frontend right now is missing some functionalities that will make it a lot easier for the user: 

-Display all current active filters being applied to the query for the user to see on the frontend
-Allow the user to selectively remove filters from the urlsearchparams after adding them
-Add debouncing protection to the enter buttons / text field
-Display "No results found" if the user enters bad query params & nothing returns. Maybe even tell them what they entered? (i. e. "No results found in area code XXXXX containing "Food", "Shelter" and X".

PR #53, opened 2020-07-14
Improve filtering frontend

andrewm898:
This addresses the issues in #46  

Here is a photo of what this looks like (working form is currently deployed)

image

-Displays all current active filters being applied to the query for the user to see on the frontend
-Allow the user to selectively remove filters from the urlsearchparams after adding them
-Add debouncing protection to the enter buttons / text field
    Right now, it just clears out the input box and requires the user to click it again and type something to enter another request. 
     This delay combined with the awaits seems to protect enough, possible todo is to add more protection in future

image

### DIFF
--- a/src/main/webapp/search_area.js
+++ b/src/main/webapp/search_area.js
@@ -41,7 +41,6 @@ class SearchArea {
     this.forOrganizationsPage = forOrganizationsPage;
     this.filterParams = new URLSearchParams();
     this.organizationObjectsList = [];
-    this.filterTagsList = [];
 
     this.zipcodeFormArea = document.createElement("div");
     this.form = document.createElement("form");
--- a/src/main/webapp/search_area.js
+++ b/src/main/webapp/search_area.js
@@ -199,7 +199,7 @@ class SearchArea {
     if (urlParamKey === "zipcode") {
       this.filterParams.delete("zipcode");
     } else {
-      if (this.filterParams.getAll("filterParam").length === 0) {
+      if (this.filterParams.getAll("filterParam").length === 1) {
         /* If only 1 filter param, delete the array */
         this.filterParams.delete("filterParam");
       } else {
--- a/src/main/webapp/search_area.js
+++ b/src/main/webapp/search_area.js
@@ -147,7 +147,6 @@ class SearchArea {
   }
 
   async setUrlParamValue(urlParamKey, urlParamValue) {   
-    
     /* New query value is not added if it is a duplicate or empty/null */
     if (this.filterParams.getAll("filterParam").includes(urlParamValue) ||
         this.filterParams.getAll("zipcode").includes(urlParamValue) ||
@@ -196,7 +195,6 @@ class SearchArea {
   }
 
   async removeFilterTag(urlParamKey, urlParamValue, filterTag) {
-
     if (urlParamKey === "zipcode") {
       this.filterParams.delete("zipcode");
     } else {
--- a/src/main/webapp/search_area.js
+++ b/src/main/webapp/search_area.js
@@ -41,6 +41,7 @@ class SearchArea {
     this.forOrganizationsPage = forOrganizationsPage;
     this.filterParams = new URLSearchParams();
     this.organizationObjectsList = [];
+    this.filterTagsList = [];
 
     this.zipcodeFormArea = document.createElement("div");
     this.form = document.createElement("form");
@@ -75,6 +76,7 @@ class SearchArea {
     this.filterInputArea.addEventListener('keypress', (e) => {
       if (e.key === 'Enter') {
         this.setUrlParamValue("filterParam", this.filterInputArea.value);
+        this.filterInputArea.value = "";
       }
     });
 
@@ -89,6 +91,10 @@ class SearchArea {
 
     this.filterInputArea.appendChild(this.filterDataList);
     this.organizationSearchArea.appendChild(this.filterInputArea);
+
+    this.activeFilterArea = document.createElement("div");
+    this.activeFilterArea.setAttribute("class", "filter-holder"); //REMOVE THIS LINE
+    this.organizationSearchArea.appendChild(this.activeFilterArea);
 
     this.organizationListArea = document.createElement("div");
     this.organizationListArea.setAttribute("id", "organization-list");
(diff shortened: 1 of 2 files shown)


src/main/webapp/search_area.js
```diff
@@ -90,6 +91,10 @@ class SearchArea {
     this.filterInputArea.appendChild(this.filterDataList);
     this.organizationSearchArea.appendChild(this.filterInputArea);
 
+    this.activeFilterArea = document.createElement("div");
+    this.activeFilterArea.setAttribute("class", "filter-holder"); //REMOVE THIS LINE
```
Sarah-Addo: Why do you have this remove this line comment?

src/main/webapp/search_area.js
```diff
@@ -75,6 +75,7 @@ class SearchArea {
     this.filterInputArea.addEventListener('keypress', (e) => {
       if (e.key === 'Enter') {
```
Sarah-Addo, line 76: Where does the user hit enter? Is there a specific enter button?

src/main/webapp/search_area.js
```diff
@@ -75,6 +75,7 @@ class SearchArea {
...
         this.setUrlParamValue("filterParam", this.filterInputArea.value);
```
Sarah-Addo, line 78: Could you add a comment about what this is doing

src/main/webapp/search_area.js
andrewm898: forgot to remove that, actually do want that line:)

src/main/webapp/search_area.js
```diff
@@ -135,13 +146,69 @@ class SearchArea {
...
+        this.removeFilterTag("zipcode", this.filterParams.get("zipcode"), document.getElementById("zipcodeTag"));
+      }
       this.filterParams.set(urlParamKey, urlParamValue);      
     } else {
       this.filterParams.append(urlParamKey, urlParamValue);      
     }
+    this.form.reset();
+    this.addFilterTag(urlParamKey, urlParamValue);
+    this.organizationObjectsList = [];
+    this.organizationListArea.innerHTML = "";
```
Sarah-Addo, line 171: .textContent

src/main/webapp/search_area.js
```diff
@@ -135,13 +146,69 @@ class SearchArea {
...
+    if (urlParamKey === "zipcode") {
+      filterTagArea.setAttribute("id", "zipcodeTag");
+    }
+
+    let filterTagLabel = document.createElement("div");
+    filterTagLabel.textContent = urlParamValue;
+    filterTagLabel.setAttribute("class", "filter-tag-label");
+    filterTagArea.appendChild(filterTagLabel);
+
+    let filterTagClose = document.createElement("div");
```
Sarah-Addo, line 189: if you want the close button to appear on the left of the label like normal then you can just move this part to be before the label part

src/main/webapp/search_area.js
andrewm898: This is a keypress for the inputArea, whenever "Enter" is hit in that input area, the input is sent

src/main/webapp/search_area.js
andrewm898: This is to clear out the divs of organizationListArea

src/main/webapp/search_area.js
andrewm898: I wanted the close button to appear to the right:)

Sarah-Addo (approved):

jnnybk (approved):
LGTM

src/main/webapp/search_area.js
```diff
@@ -135,13 +147,69 @@ class SearchArea {
...
+  async setUrlParamValue(urlParamKey, urlParamValue) {   
+    /* New query value is not added if it is a duplicate or empty/null */
+    if (this.filterParams.getAll("filterParam").includes(urlParamValue) ||
+        this.filterParams.getAll("zipcode").includes(urlParamValue) ||
+        (urlParamValue === null) || (urlParamValue.trim() === "")) {
+      return;
+    }
+    
+    /* if the param is a zipcode, remove tag of any existing one & set new one*/ 
     if (urlParamKey === "zipcode") {
```
belraquib, line 159: Being nitpicky for future you:

This looks good in general, but consider if you want a new param that you want to have this single-value-only type semantics.

What if you had a Set that described which parameters you wanted as singleton

Set<String> singletonParameters = ....  right now just "zipcode"

Then a lot of these functions will read:

if (singletonParameters.contains(urlParamKey)) {  ... }

Then you can add new singleton parameters easily.

src/main/webapp/search_area.js
```diff
@@ -135,13 +147,69 @@ class SearchArea {
...
+      if (this.filterParams.get("zipcode")) {
```
belraquib, line 160: use urlParamKey here, it will be easier to add to the filterParam set in the future.

src/main/webapp/search_area.js
```diff
@@ -135,13 +147,69 @@ class SearchArea {
...
+        /* If there is a zipcode being displayed, remove its tag so both aren't displayed */
+        this.removeFilterTag("zipcode", this.filterParams.get("zipcode"), document.getElementById("zipcodeTag"));
```
belraquib, line 162: will all singleton key elements be called `<keyname>Tag` ?  If so, construct the name dynamically.

src/main/webapp/search_area.js
```diff
@@ -135,13 +147,69 @@ class SearchArea {
...
+    await this.getListOfOrganizations();
+    this.renderListOfOrganizations();
+  }
+
+  addFilterTag(urlParamKey, urlParamValue) {
+    let filterTagArea = document.createElement("div");
+    filterTagArea.setAttribute("class", "filter-tag-area");
+    /* ID is given to zipcode tag so it can be removed if new one is added */
+    if (urlParamKey === "zipcode") {
+      filterTagArea.setAttribute("id", "zipcodeTag");
```
belraquib, line 181: if you have a static Set of singleton keys you can generalize zip code here.

(I can imagine a filter by State, for example)

belraquib (approved):
Feel free to punt on the singleton work in another PR.

PR #56, opened 2020-07-15
Create front-end part of adding hours to organization during registration

Sarah-Addo:
Screenshot: https://screenshot.googleplex.com/YDig89HroPn


src/main/webapp/open_hours.js
```diff
@@ -0,0 +1,99 @@
...
+
+class TimeOption {
+  constructor(day, forRegistration, organization) {
+    this.day = day;
+    this.forRegistration = forRegistration;
+    this.organization = organization;
+    this.optionArea = document.getElementById("hours-option-area");
+    if(organization != null) {
+      if(day == "Monday") {
+        this.organizationIsOpen = organization.mondayIsOpen;
```
jnnybk: I think it could be confusing to have organizationIsOpen when it's holding open/close status for just one day!

jnnybk (approved):

src/main/webapp/open_hours.js
Sarah-Addo: I understand what you mean. These objects are constructed for unique days so it is trying to refer to if the organization is open for the day it was created for but I was thinking I might need this for when I use it for the edit page but I think i figured out another way so i removed it!

src/main/webapp/open_hours.js
```diff
@@ -0,0 +1,82 @@
+document.addEventListener("DOMContentLoaded", async function() {
```
belraquib, line 15: Add a top level comment to what this class should do, what it expects to be present in the DOM, etc.

src/main/webapp/open_hours.js
```diff
@@ -0,0 +1,82 @@
...
+  if(document.getElementById("registration-form-area")) {
+  const mondayTimeOption = new TimeOption("Monday", true, null);
```
belraquib: are these used anywhere?

belraquib: ah, i see: these register objects that themselves have listeners on the DOM.

src/main/webapp/open_hours.js
```diff
@@ -0,0 +1,82 @@
...
+    this.forRegistration = forRegistration;
+    this.organization = organization;
+    this.optionArea = document.getElementById("hours-option-area");
+
+    this.dayOptionArea = document.createElement("div");
+    this.dayOptionArea.classList.add("day-option-area");
+    this.timeInputArea = document.createElement("div");
+
+    this.dayLabel = document.createElement("label");
+    this.dayLabel.textContent = `${this.day}: `;
```
belraquib: usually when I see space in a text context, that suggests to me that you are trying to manually space something that should be spaced via css padding. Please check how this renders without the spaces?

belraquib (changes requested):

andrewm898 (approved):

belraquib (approved):

PR #47, opened 2020-07-13
add function stubs for organization updater invited moderator changes

Sarah-Addo:

PR #59, opened 2020-07-16
Implement Moderator logic

jnnybk:
Definitions:
- _Moderator_ is a User who oversees and maintains an Organization.
- _InvitedModerators_ list is a property of a Distributor (type of Organization) that holds the emails of Moderators - who have not logged into this application - for that specific Distributor.
- _ModeratorList_ is a property of a Distributor that holds the IDs of Moderators - who have logged in.

NOTE: The reason for storing emails and IDs separately is because Users API does not allow emails to identify IDs, Therefore, when a User is not logged in, we cannot access their IDs and when the User logs in, we then can update their ID based on a matching email address.

This PR accomplishes the following:
1) Implement updating Moderator from "invitedModerators" list to "moderatorList".
2) Add moderatingOrgs attribute of type ArrayList of Organization entities to GivrUser class:
- This was chosen over ArrayList of long IDs because OrganizationUpdater.java uses Organization entity as the parameter when updating Moderator (1).
3) Implement querying to Datastore for Distributor's moderatorList via a CompositeFilter (operation OR) to check both moderatorList and invitedModerators list of Distributor's properties.
4) Use PR #47 changes to complete this feature.


jnnybk:
TODO: Implement adding User to "invitedModeratorList" in AuthenticateServlet if they are invited as a moderator but have not logged in.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,6 +106,29 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
...
+
+    this.moderatingOrgs = new ArrayList<Entity>();
+
+    Query query = new Query("Distributor").addSort("creationTimeStampMillis", SortDirection.DESCENDING);
+
+    // TODO(): Check in invitedModeratorList as well.
+
+    Filter filter = null;
+    try {
+      filter = new FilterPredicate("moderatorList", FilterOperator.IN, this.email);
```
Sarah-Addo: shouldn't you be looking for user id's and not emails since moderator list contains their ids? Also what happens if they are an invited moderator but not in the moderator list yet

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -130,41 +128,33 @@ private void setOrganizationProperty(String propertyKey, String formValue) {
...
+        Set<String> invitedModerators = new HashSet((HashSet)this.entity.getProperty("invitedModerators"));
+        invitedModerators.add(email);
+      } else {
+        userIds.add(newUser.getUserId());
+      }
     }
     return userIds;
   }
 
-  private void updateNonFormProperties(GivrUser user, boolean forRegistration, EmbeddedEntity historyUpdate) {
```
Sarah-Addo, line 141: No do not delete this method it serves a different purpose

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -98,10 +100,6 @@ public void updateOrganization(HttpServletRequest request, GivrUser user, boolea
 
       setOrganizationProperty(propertyKey, formValue);
     }
-
```
Sarah-Addo, line 101: again do not delete this

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -53,14 +57,12 @@ public void updateOrganization(HttpServletRequest request, GivrUser user, boolea
...
     properties.put("org-email", "orgEmail");
     properties.put("org-street-address", "orgStreetAddress");
-    properties.put("org-zip-code", "orgZipCode");
     properties.put("org-phone-num", "orgPhoneNum");
     properties.put("org-url", "orgUrl");
     properties.put("org-description","orgDescription");
     properties.put("approval", "isApproved");
     properties.put("moderator-list", "moderatorList");
 
-    // Updates entity properties from form
```
andrewm898: why delete this comment?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -53,14 +57,12 @@ public void updateOrganization(HttpServletRequest request, GivrUser user, boolea
     properties.put("org-name", "orgName");
     properties.put("org-email", "orgEmail");
     properties.put("org-street-address", "orgStreetAddress");
-    properties.put("org-zip-code", "orgZipCode");
```
andrewm898: this cant be deleted

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: fixed by using Composite Filters!

Sarah-Addo (approved):

andrewm898 (approved):

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,6 +108,26 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
     datastore.put(entity);
   }
 
+  public void setModeratingOrgs() {
```
mickrobk, line 126: We should avoid making users of this class call an extra function for the data to be valid. (Encapsulate this state!) Basically: make it impossible for the class to have incorrect data once it's been constructed

Eg: if I read in some other  other part of the codebase:
void MyFunc(GivrUser user) {
  if (!user.moderatingOrgs.contains("donor1") {
    return;
  }
}

Did that code forget to call "setModeratingOrgs"? Is moderatingOrgs empty? If we add a call there, are we calling it twice?

It's much more important for things to be correct vs avoiding duplicate work. Can you instead mark this function private, and call it in the constructor? If this is truly super expensive to do; there are ways we can structure this code so it's both encapsulated and lazy. Happy to chat about options, but I don't think that's the case here!

src/main/webapp/user.js
```diff
@@ -63,7 +62,7 @@ class User {
       organizationsLink.setAttribute("href", "organizations.html");
       organizationsLink.textContent = "Organizations";
       navBar.appendChild(organizationsLink);
-    } else if (isModerator) {
+    } else if (this.isModerator) {
```
mickrobk, line 123: Nice!

mickrobk (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: This makes sense! I will make the appropriate changes. Thank you!

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -42,13 +45,26 @@
...
     this.isMaintainer = isMaintainer;
     this.isLoggedIn = isLoggedIn;
     this.url = url;
     this.email = email;
+    this.moderatingOrgs = moderatingOrgs;
+
+    if (!this.isMaintainer) {
+        setModeratingOrgs();
+        ArrayList<Entity> orgList = getModeratingOrgs();
+        if (orgList.size() > 0) {
```
mickrobk: Do you need this if statement?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -42,13 +45,26 @@
...
+          for (int i = 0; i < orgList.size(); i++) {
+            OrganizationUpdater organizationUpdater = new OrganizationUpdater(orgList.get(i));
+            organizationUpdater.updateInvitedModerator(this);
```
mickrobk: Avoiding non-obvious side effects is a big thing in writing maintainable code. In general:
 - the constructor + any getters should be read-only. (Don't make changes to other classes or datastores)
 - Things that do mutate state can be called "updateXYZ" / "setXYZ" / "applyXYZ"; basically an action verb

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -42,13 +45,26 @@
...
+  public GivrUser(String id, boolean isMaintainer, boolean isLoggedIn, String url, String email, ArrayList<Entity> moderatingOrgs) {
     this.id = id;
     this.isMaintainer = isMaintainer;
     this.isLoggedIn = isLoggedIn;
     this.url = url;
     this.email = email;
+    this.moderatingOrgs = moderatingOrgs;
+
+    if (!this.isMaintainer) {
+        setModeratingOrgs();
```
mickrobk: I'm not sure I understand what's happening here. Looks like you set the class orgs to the passed in arg, then overwrite it with a db read, then write changes to the org db?

Why do you need lines 58-67?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -44,7 +44,12 @@ public void updateOrganization(HttpServletRequest request, GivrUser user, boolea
     Set<String> requiresModerator = new HashSet<String>();
     Map<String, String> properties = new HashMap<String,String>();
     boolean isMaintainer = user.isMaintainer();
-    boolean isModerator = user.isModerator(this.entity.getKey().getId());
+    boolean isModerator = false;
+
+    // Gets moderator status from checking its moderatingOrgs size.
+    if (user.getModeratingOrgs().size() > 0) {
+      isModerator = true;
+    }
```
mickrobk: Could you encapsulate this logic in user? What if you had a function: User.IsModeratorOfAnyOrg?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -130,10 +135,17 @@ private void setOrganizationProperty(String propertyKey, String formValue) {
...
-      userIds.add(newUser.getUserId());
+      String userId = newUser.getUserId();
+      // If failed will return "" (?) and user will be added to the invited moderator property 
+      // and not the offical list
+      if(userId.equals("")) {
+        Set<String> invitedModerators = new HashSet((HashSet)this.entity.getProperty("invitedModerators"));
+        invitedModerators.add(email);
+      } else {
+        userIds.add(newUser.getUserId());
+      }
```
mickrobk: Per other comment -- take care to separate out the mutation functions from the read functions, and to make sure what's happening is obvious from just the name of the function. Eg reading "translateEmailsToIds": I would not guess that the emails are added to the moderator list.

Can you decompose this into a set of blocks that have obvious names and no side effects?

Ex to add a moderator:
array emails[]
for (email in emails) 
  user = UserFromEmail
  org.addmoderator(user)

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -145,6 +181,6 @@ public static GivrUser getCurrentLoggedInUser() {
     if (isUserLoggedIn) {
       return getUserById(userService.getCurrentUser().getUserId());
     }
-    return new GivrUser("" /* userId */, false /* isMaintainer */, false /* isLoggedIn */, url /* loginURL */, "" /* userEmail */);
+    return new GivrUser("" /* userId */, false /* isMaintainer */, false /* isLoggedIn */, url /* loginURL */, "" /* userEmail */, new ArrayList<Entity>());
```
mickrobk: Can you avoid constructing an invalid givruser? (Eg the moderator list here is wrong) What if you changed "private void setModeratingOrgs" to "private array readModeratingOrgs"?

mickrobk (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: No! I removed it! Thank you!

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: Noted! I'm now wondering whether it would be okay to just call the setModeratingOrgs() within GivrUser's constructor, and within setModeratingOrgs(), update each Organization (that the GivrUser is a moderator of) using OrganizationUpdater. Somehow, my code needs to accomplish updating Organization entities in the Datastore.

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: I removed passing moderatingOrgs into the GivrUser constructor, and only called setModeratingOrgs() from the constructor. 

I moved the code from 58-67 to setModeratingOrgs(). This code does the updating to the Organization entities based on the OrganizationUpdater. The method that updates to the Datastore in the OrganizationUpdater takes in a GivrUser object, in order to update the appropriate Organization for a specific user.

src/main/java/com/google/sps/data/OrganizationUpdater.java
jnnybk: I refactored this code and added a few comments!

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: I removed the moderatingOrgs from the constructor, so it should now be setting the correct moderatingOrgs.

src/main/java/com/google/sps/data/GivrUser.java
mickrobk: Why does it need to do that in the constructor though? In particular, the new function: "setModeratingOrgsAndUpdateOrgs" has a great name, as it's obvious what it does. But: why is that the function that exists instead of "setModeratingOrgs" + "UpdateOrgs"

Should "updateOrgs" be a function on orgupdater instead of giveruser?

src/main/java/com/google/sps/data/GivrUser.java
mickrobk: This is much clearer now, thanks!

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -127,15 +133,23 @@ private void setOrganizationProperty(String propertyKey, String formValue) {
...
       GivrUser newUser = GivrUser.getUserByEmail(email);
-      userIds.add(newUser.getUserId());
+      String userId = newUser.getUserId();
+      // UserId can equal "" if that user has never logged in. User email will be added to the invitedModerators list, and not the moderatorList.
+      if (userId.equals("")) {
+        if (this.entity.getProperty("invitedModerators") == null) {
+          invitedModeratorEmails = new ArrayList<String>();
+        } else {
+          invitedModeratorEmails = (ArrayList) this.entity.getProperty("invitedModerators");
+        }
```
mickrobk, line 182: The way java arguments work (references by value), this is not modifying the "invitedModeratorEmails" that was passed in!

Ex:
public class Experiment {
    public static void main(String args[]) {
      String s = "In";
      foo(s);
      System.out.println(s);
    }
    public static void foo(String s) {
        s = "Out";
    }
}

Output:
"In"

src/main/webapp/user.js
```diff
@@ -29,8 +29,12 @@ class User {
     if (loginData.isLoggedIn) {
       this.isMaintainer = loginData.isMaintainer;
 
-      // TODO(): add isModerator check
-      this.rebuildNavBar(/*isModerator*/ false);
+      if (loginData.moderatingOrgs.length === 0) {
+        this.isModerator = false;
+      } else {
+        this.isModerator = true;
+      }
```
mickrobk: this.isModerator = loginData.moderatingOrgs.length >= 1

mickrobk (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: Yes, we do have an updateInvitedModerator method in OrgUpdater which is called in the setModeratingOrgsAndUpdateOrgs.

The reason why we need update to be invoked within GivrUser is because each Organization that the user is a moderator of needs: 1) remove the user's email from the invitedModerators list of that specific Organization, and 2) add the user's ID to the moderatorList of that Organization.

I will refactor the setModeratingOrgs and UpdateOrgs to be separate methods, but I'm not sure if I should be calling updateOrgs within the GivrUser's constructor. Somehow, this needs to be called whenever a user logs in, and this could be in the getCurrentLoggedInUser().

mickrobk: Would everything work for now if we skipped 1,2?

src/main/java/com/google/sps/data/OrganizationUpdater.java
jnnybk: I refactored this to call two separate functions to find invited moderators vs active moderators and return the appropriate lists!

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: Just responding to this comment to document that we have discussed and came to a solution for this issue: Call setting moderating orgs and update organizations (invited moderator list -> moderator list) within Authenticate Servlet, as every user (both logged in & not logged in) calls it.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,6 +115,38 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
     datastore.put(entity);
   }
 
+  public void updateModeratingOrgs() {
```
belraquib, line 118: Add a comment that this updates the organizations (not reading the moderating organizations for the user which is how I approached it as first.) In general, all public functions should have some documentation.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,6 +115,38 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
...
+    for (int i = 0; i < this.moderatingOrgs.size(); i++) {
+      OrganizationUpdater organizationUpdater = new OrganizationUpdater(this.moderatingOrgs.get(i));
+      organizationUpdater.updateInvitedModerator(this);
+    }
+  }
+
+  public void setModeratingOrgs() {
+    // There is no need to set moderatingOrgs for maintainers.
```
belraquib: Are you sure? Does this mean that any moderator that happens to be a maintainer never moves off the invited list?

belraquib (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: That is a great catch! I removed the check for the maintainer status.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,6 +116,36 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
...
+  // Updates Organization entities in Datastore by removing user's email from invitedModerators list and adding user's ID in moderatorlist.
+  public void updateModeratingOrgs() {
+    for (int i = 0; i < this.moderatingOrgs.size(); i++) {
+      OrganizationUpdater organizationUpdater = new OrganizationUpdater(this.moderatingOrgs.get(i));
+      organizationUpdater.updateInvitedModerator(this);
+    }
+  }
+
+  // Properly sets user's moderatingOrgs based on results from querying to Datastore for Organization entity's moderatorList and invitedModerators list.
+  public void setModeratingOrgs() {
```
belraquib, line 126: Note in the comments that you don't update the datastore here?

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -67,9 +73,15 @@ public boolean isLoggedIn() {
     return this.isLoggedIn;
   }
 
-  // TODO(): get correct moderator status for organization
-  public boolean isModerator(long organizationId) {
-      return false;
+  public ArrayList<Entity> getModeratingOrgs() {
+    return this.moderatingOrgs;
```
belraquib, line 78: I want you to think about how you can make sure getModeratingOrgs will always return the right thing. Right now the user may not have called setModeratingOrgs yet.

belraquib (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: My idea is to call setModeratingOrgs() before checking getModeratingOrgs().size(). Is there a better solution?

belraquib: Yup. Consider: do you want to do it here instead of in isModeratorOfAnyOrg?

jnnybk: yes! That would be better, thank you! :D

src/main/java/com/google/sps/servlets/AuthenticateServlet.java
```diff
@@ -79,6 +81,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse  response) thr
     Gson gson = new Gson();
 
     GivrUser user = GivrUser.getCurrentLoggedInUser();
+    // This servlet will always be called for logged in and non logged in users.
+    if (user.isModeratorOfAnyOrg()) {
```
jnnybk, line 85: I removed user.setModeratingOrgs() that was originally on line 84 because getModeratingOrgs() calls setModeratingorgs().

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -67,9 +73,16 @@ public boolean isLoggedIn() {
...
-  // TODO(): get correct moderator status for organization
-  public boolean isModerator(long organizationId) {
-      return false;
+  public ArrayList<Entity> getModeratingOrgs() {
+    setModeratingOrgs(); // Makes sure that getModeratingOrgs is returning updated information.
+    return this.moderatingOrgs;
+  }
+
+  public boolean isModeratorOfAnyOrg() {
+    if (getModeratingOrgs().size() > 0) {
```
belraquib: consider the following:

if (statement evaluates to true) {
  return true;
}
return false

vs

return statement

belraquib (approved):

PR #61, opened 2020-07-17
Filter frontend param update

andrewm898:
This changes the filtering frontend to have an "add filter" button that the user can use to add a new filter, first by choosing a type / thing to filter by, and then by typing the actual keyword. All of this filtering frontend functionality is in the FilterTagArea class.

Note: This needs to be merged after #62, accidentally opened them in the wrong order

Here is the sequence of how this new filtering frontend works:

The user clicks "Add Filter", a new filter tag pops up where they select what they want to filter by
image

After selecting what they want to filter by, in this case "Organization Name", the user enters a word for the type of filter they selected
image

That filter is then processed and the results are adjusted (This backend is in a separate PR)
image

Notes / TODO's: 
- The user can enter any amount of categories, org names, or addresses right now, the only one that is limited is zip codes
- TODO: if the user enters a filter category that is not recognized, it is currently ignored. A TODO is to notify the user of this


src/main/webapp/search_area.js
```diff
@@ -166,50 +139,7 @@ class SearchArea {
...
     }
     this.form.reset();
-    this.addFilterTag(urlParamKey, urlParamValue);
-    this.organizationObjectsList = [];
-    this.organizationListArea.innerHTML = "";
-    await this.getListOfOrganizations();
-    this.renderListOfOrganizations();
-  }
-
-  addFilterTag(urlParamKey, urlParamValue) {
```
andrewm898, line 177: note: addFilterTag and removeFilterTag are now parts of the FilterTagArea class

src/main/webapp/filter_tag_area.js
```diff
@@ -0,0 +1,140 @@
...
+    this.activeFilterArea = document.createElement("div");
+    this.activeFilterArea.setAttribute("class", "filter-holder");
+
+    /* filterEntryArea is displayed when the user clicks add filter. It first asks for a type of filter, 
+       then asks the user to enter the keyword for that filter type */
+    this.filterEntryArea = document.createElement("div");
+    this.filterEntryArea.setAttribute("class", "filter-tag-area");
+    this.filterEntryArea.setAttribute("id", "filter-entry");
+
+    this.filterEntryClose = document.createElement("div");
```
Sarah-Addo: You're gonna run into a merge conflict because i rearranged the order a little bit and changed some styling but you can just keep everything in this pr and I can redo it after you merge

jnnybk (approved):

Sarah-Addo (approved):

src/main/webapp/filter_tag_area.js
```diff
@@ -0,0 +1,143 @@
...
+      this.activeFilterArea.removeChild(this.filterEntryArea)
+    });
+
+    /* The user selects the type of property they want to filter by in filterTypeInput*/
+    this.filterTypeInput = document.createElement("input");
+    this.filterTypeInput.setAttribute("list", "filter-datalist");
+    this.filterTypeInput.setAttribute("class", "filter-input-area");
+    this.filterTypeInput.setAttribute("placeholder", "Filter by:");
+    this.filterTypeInput.addEventListener('keypress', (e) => {
+      if (e.key === 'Enter') {
```
mickrobk: do you want to listen for enter or for "onselect"/"onchange"?

src/main/webapp/filter_tag_area.js
```diff
@@ -0,0 +1,143 @@
...
+    for (const optionKey of this.optionMap.keys()) {
+      const option = document.createElement("option");
+      option.value = optionKey;
+      this.filterDataList.appendChild(option);
+    }
+
+    /* After the type has been chosen, the actual filter param is entered here */
+    this.filterParamInput = document.createElement("input");
+    this.filterParamInput.setAttribute("class", "filter-input-area");
+    this.filterParamLabel = document.createElement("div");
```
mickrobk: Should this be a label instead of a div?

src/main/webapp/filter_tag_area.js
```diff
@@ -0,0 +1,143 @@
...
+    this.addFilterButton.addEventListener('click', () => {
+      /* If the user is on the second part of entering a filter, add filter is disabled */
+      if (!this.filterEntryArea.contains(this.filterParamInput)) {
+        this.filterTypeInput.appendChild(this.filterDataList);
+        this.filterEntryArea.appendChild(this.filterEntryClose);
+        this.filterEntryArea.appendChild(this.filterTypeInput);
+        this.activeFilterArea.appendChild(this.filterEntryArea);
+      }
+    });
+    this.activeFilterArea.appendChild(this.addFilterButton);
```
mickrobk: Is this class meant to be the area that holds all the filters, a single filter, or both? Would the code be simpler if you decomposed it into the 2 separate concepts?

src/main/webapp/filter_tag_area.js
```diff
@@ -0,0 +1,143 @@
...
+        let filterArray = this.parentSearchArea.filterParams.getAll(urlParamKey);
+        filterArray.splice(filterArray.indexOf(urlParamValue), 1);
+        this.parentSearchArea.filterParams.set(urlParamKey, filterArray);
+      }
+    }
+    this.activeFilterArea.removeChild(filterTag);
+    this.parentSearchArea.organizationObjectsList = [];
+    this.parentSearchArea.organizationListArea.innerHTML = "";
+    await this.parentSearchArea.getListOfOrganizations();
+    this.parentSearchArea.renderListOfOrganizations();
```
mickrobk: Why update the org list here, but not in addFilterTag (+elsewhere)?

mickrobk (changes requested):

src/main/webapp/filter_tag_area.js
andrewm898: I made it like this so the user can enter their own input and then hit enter when they are satisfied with it

src/main/webapp/filter_tag_area.js
andrewm898: This class holds the add filter button, the tag that pops up when you click add filter, and then whatever filter is being applied. Considering how connected these concepts are, I think it's definitely simplest to encapsulate them all into the same class.

src/main/webapp/filter_tag_area.js
mickrobk: Ok, rephrasing: we're doing it this way so they can explicitly change focus without committing?

src/main/webapp/filter_tag_area.js
mickrobk: I'm not sure I agree that it makes it simpler! Encapsulating code (even if there's more of it) results in less that needs to be understood, as things get pulled into helper functions, documented, and decoupled. 

I'd go so far as to say that teasing apart where you can break apart related concepts is one of the most powerful things you can do to simplify a design.

src/main/webapp/filter_tag_area.js
andrewm898: removeFilterTag is the function that is called when the "x" on a filter tag is clicked, so it clears the organization list and then re-queries for it with the new set of filters. addFilterTag doesn't need to worry about updating the org list because it is only designed to add the tag to the dom, since when new url params are added getting the new list is handled by the url search params function in search_area.js

mickrobk: Rephrased: updating the org list and removing a filter seem like unrelated tasks, tied together only by the context in which they're used.

Can you break these apart into their own functions? (And reuse the refreshOrgList function wherever applicable!)

andrewm898: I was originally updating the org list in the setUrlParamValue function after addFilterTag was called, but I agree that it makes it the code a little bit more readable to have addFilterTag and removeFilterTag do the same thing, so the new commit changes it to do it that way

src/main/webapp/filter_tag_area.js
andrewm898: The FilterTagArea class now holds the add filter button / text box, and the tags themselves are represented by the FilterTag class.

mickrobk: If you had a library with the following interface:
```
class FilterEntry {
  // Constructs a new filter entry row.
  // Starts with:
  //   - a text area to enter the filter field
  //   - hasFilterField == false, hasFilterValue == false
  // After the filter field is entered:
  //   - there is a second text area to enter the filter value
  //   - hasFilterField == true, hasFilterValue == false
  // After the filter value is entered:
  //   - hasFilterField == true, hasFilterValue == true
  // There is also a close button in the entry row. When clicked
  // the filter row will be removed, and onRemoveCallback(this) is called
  constructor(div parentElement, onRemoveCallback)

  bool hasFilterField()
  string getFilterField()

  bool hasFilterValue()
  string getFilterValue()
}
```

Would it make the implementation of FilterTagArea easier?

andrewm898: That makes sense as a way to break up some of the tasks being done in the FilterTagArea constructor- I think it would make the implementation more straighforward

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,105 @@
...
+        this.filterFieldInput.removeChild(this.filterDataList);
+        this.filterEntryArea.removeChild(this.filterFieldInput);
+        this.filterEntryArea.appendChild(this.filterParamInput);
+        this.filterEntryArea.appendChild(this.filterParamLabel);
+      }
+    });
+
+    this.filterDataList = document.createElement("datalist");
+    this.filterDataList.setAttribute("id", "filter-datalist");
+    this.optionMap = new Map();
```
belraquib: consider doing:

new Map(
  ["Organization Name", "orgNames"], 
  ["Address", "orgStreetAddress"], 
  [....]);

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,105 @@
...
+    this.filterEntryArea.setAttribute("class", "filter-tag-area");
+    this.filterEntryArea.setAttribute("id", "filter-entry");
+
+    /* The user selects the type of property they want to filter by in filterFieldInput*/
+    this.filterFieldInput = document.createElement("input");
+    this.filterFieldInput.setAttribute("list", "filter-datalist");
+    this.filterFieldInput.setAttribute("class", "filter-input-area");
+    this.filterFieldInput.setAttribute("placeholder", "Filter by:");
+    this.filterFieldInput.addEventListener('keypress', (e) => {
+      if (e.key === 'Enter') {
```
belraquib: nit : i find the following structure more readable (here and elsewhere)

(e) -> {
  if (e.key !== 'Enter') {
    return;
  }
  // e.key was enter, do the rest
}


This makes it so I (the reader) don't have to remember the scope of the code inside the curly brace. In other words, as written I would read this code as:

if the key is enter then i do the following, x, y, and z.

In the change I proposed, you would read it as

if the key is not enter, there is nothing to do, return
do x, y, z

If you think of code as branches off a tree what you are trying to do is keep the branch factor uncomplicated and have as many things in the "trunk" as possible

In this case, it's not as important, but it does become much more important if the stuff in side the if gets more complicated (has other branches).

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,105 @@
...
+
+    /* After the type has been chosen, the actual filter param is entered here */
+    this.filterParamInput = document.createElement("input");
+    this.filterParamInput.setAttribute("class", "filter-input-area");
+    this.filterParamLabel = document.createElement("label");
+    this.filterParamLabel.setAttribute("class", "filter-tag-label");
+    this.filterParamInput.addEventListener('keypress', (e) => {
+      if (e.key === 'Enter') {
+        /* If the user enters an unsupported type, it is ignored on the servlet side */
+        console.log(this.optionMap.get(this.filterFieldInput.value));
```
belraquib: remove log?

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,105 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/* 
+ * FilterEntry displays an area for the user to first specify the field they want to filter by,
+ * and then the keyword they want to apply to that filter field
+ */
```
belraquib, line 21: You should describe the events you fire (and what causes those events to fire) in the comments here.

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,105 @@
...
+          detail : {
+            urlParamKey: this.optionMap.get(this.filterFieldInput.value),
+            urlParamValue: this.filterParamInput.value
+          }
+        }));
+        this.filterParamInput.value = "";
+        this.filterFieldInput.value = "";
+        this.filterEntryArea.removeChild(this.filterParamInput);
+        this.filterEntryArea.removeChild(this.filterParamLabel);
+        this.filterEntryArea.dispatchEvent(this.closeEvent);
```
belraquib: instead of this.closeEvent, just create a new one each time 

new Event('onRemove')

src/main/webapp/search_area.js
```diff
@@ -148,14 +131,16 @@ class SearchArea {
     this.organizationObjectsList = await response.json();
   }
 
-  async setUrlParamValue(urlParamKey, urlParamValue) {   
+  async setUrlParamValue(urlParamKey, urlParamValue) {
+    console.log("FLKJGKHHFKJSDFHGKJHFSLKUHSDGHFDJGFSDJHGSDKHFGKSDJHGFVDGSFKJDGVSFKJHSDBVKDGFSKHFBDFSKJHGB");
```
belraquib: remove logs

belraquib (changes requested):

src/main/webapp/filter_entry.js
andrewm898: my bad, forgot to remove those

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,107 @@
...
+    /* The user selects the type of property they want to filter by in filterFieldInput*/
+    this.filterFieldInput = document.createElement("input");
+    this.filterFieldInput.setAttribute("list", "filter-datalist");
+    this.filterFieldInput.setAttribute("class", "filter-input-area");
+    this.filterFieldInput.setAttribute("placeholder", "Filter by:");
+    this.filterFieldInput.addEventListener('keypress', (e) => {
+      if (e.key !== 'Enter') {
+        return
+      }
+        this.filterParamLabel.textContent = `${this.filterFieldInput.value}:`;
```
belraquib: fix indentation :)

src/main/webapp/filter_entry.js
```diff
@@ -0,0 +1,107 @@
...
+      this.filterDataList.appendChild(option);
+    }
+
+    /* After the type has been chosen, the actual filter param is entered here */
+    this.filterParamInput = document.createElement("input");
+    this.filterParamInput.setAttribute("class", "filter-input-area");
+    this.filterParamLabel = document.createElement("label");
+    this.filterParamLabel.setAttribute("class", "filter-tag-label");
+    this.filterParamInput.addEventListener('keypress', (e) => {
+      if (e.key === 'Enter') {
```
belraquib: same change as before.

src/main/webapp/index.html
```diff
@@ -1,14 +1,17 @@
...
-<head>
-	<meta charset="UTF-8">
-	<title>Givr</title>
-	<link rel="stylesheet" href="style.css">
-	<script src="user.js"></script>
-	<script src="organization.js"></script>
-	<script src="search_area.js"></script>
-</head>
+  <head>
+	  <meta charset="UTF-8">
```
belraquib: check your lint here

belraquib (approved):

PR #62, opened 2020-07-17
Filtering backend improvements

andrewm898:
This adds a resourceCategories field in the registration/organization updating sections, and adjusts ListOrganizationServlet to handle the new format of filter parameters from the frontend

TODO: modify edit org servlet to handle resource categories as well


src/main/webapp/style.css
```diff
@@ -64,12 +64,12 @@
   height: 30px;
 }
```
andrewm898: ignore all changes in style.css- this branch has the most recent form of master

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -59,6 +59,7 @@ public void updateOrganization(HttpServletRequest request, GivrUser user, boolea
     properties.put("org-description","orgDescription");
     properties.put("approval", "isApproved");
     properties.put("moderator-list", "moderatorList");
+    properties.put("org-resource-category", "resourceCategories");
```
Sarah-Addo, line 62: you only made sure that adding an organization works but you also must modify the edit page too i.e now only have that area but also prefill the values in the form. Can you also fix this for zip code because the edit page is also failing because of that

Sarah-Addo: please test it when you're done too haha

andrewm898: I left the updating of the edit page as a TODO, because on master/my branches it still looks like the edit page is inaccessible because of those bugs from before where it just shows the help near me and register orgs page. How can I access the page that shows the organizations so I can edit them?

andrewm898: resolved and tested

jnnybk (approved):
LGTM

Sarah-Addo (approved):

mickrobk (approved):
The work in simplifying the filtering stuff is paying off!

PR #63, opened 2020-07-17
 add open hours backend

### LINKED ISSUE #51: Implement Open Hours tracking
Implement a data structure to represent the hours an organization is open on each day of the week, and display this info with the rest of the popup info

Sarah-Addo:
Popup Screenshot: https://screenshot.googleplex.com/N5poY7xj249
Edit page screenshot: https://screenshot.googleplex.com/LUsOMp8HssD


andrewm898 (approved):

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -29,6 +29,16 @@
...
 
+  enum DayOfWeek {
+  Monday,
+  Tuesday,
+  Wednesday,
+  Thursday,
+  Friday,
+  Saturday,
+  Sunday
+  }
```
mickrobk: Can you take a formatting pass?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -167,4 +187,39 @@ private void updateNonFormProperties(GivrUser user, boolean forRegistration, Emb
...
+      }
+      hoursOpen.add(dayOption);
+    }
+
+    this.entity.setProperty("orgHoursOpen", hoursOpen);
+  }
+
+  private ArrayList<EmbeddedEntity> createFromToPairs(ArrayList<String> dayOptionFromTimes, ArrayList<String> dayOptionToTimes) {
+    ArrayList<EmbeddedEntity> pairs = new ArrayList<EmbeddedEntity>();
+    for(int i = 0; i < dayOptionFromTimes.size(); i++) {
```
mickrobk, line 236: Check that dayOptionToTimes.size() == FromTimes.size()?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -167,4 +187,39 @@ private void updateNonFormProperties(GivrUser user, boolean forRegistration, Emb
     
     this.entity.setProperty("lastEditTimeStampMillis", millisecondSinceEpoch);
   }
+
+  private void updateOpenHoursProperty(HttpServletRequest request) {
+    ArrayList<EmbeddedEntity> hoursOpen = new ArrayList<EmbeddedEntity>();
+    ArrayList<String> dayOptionFromTimes = new ArrayList<String>();
+    ArrayList<String> dayOptionToTimes = new ArrayList<String>();
```
mickrobk: It looks like these are only used inside the if block, inside the for block. Can you move these declarations into the innermost scope s.t. they are better encapsulated?

src/main/java/com/google/sps/data/OrganizationUpdater.java
```diff
@@ -167,4 +187,39 @@ private void updateNonFormProperties(GivrUser user, boolean forRegistration, Emb
...
+
+    this.entity.setProperty("orgHoursOpen", hoursOpen);
+  }
+
+  private ArrayList<EmbeddedEntity> createFromToPairs(ArrayList<String> dayOptionFromTimes, ArrayList<String> dayOptionToTimes) {
+    ArrayList<EmbeddedEntity> pairs = new ArrayList<EmbeddedEntity>();
+    for(int i = 0; i < dayOptionFromTimes.size(); i++) {
+      EmbeddedEntity fromToPair = new EmbeddedEntity();
+      fromToPair.setProperty("from", dayOptionFromTimes.get(i));
+      fromToPair.setProperty("to", dayOptionToTimes.get(i));
```
mickrobk, line 239: Do you want to do any validation here? What if from > to, or one of the values is not a time? (

src/main/webapp/organization.js
```diff
@@ -301,6 +308,70 @@ class Organization {
...
+      const numPairs = day.propertyMap.fromToPairs.value.length;
+      for (let i = 0; i < numPairs; i++) {
+        let from = this.parseTime(day.propertyMap.fromToPairs.value[i].propertyMap.from);
+        let to = this.parseTime(day.propertyMap.fromToPairs.value[i].propertyMap.to);
+        if (numPairs - 1 != i) {
+          fromToString = `${from} - ${to}, \n`;
+        }
+        else {
+          fromToString = `${from} -  ${to} \n`;
+        }
```
mickrobk: Instead of duplicating the thing you want to print with a small variation, what if you did something more like:
fromToString = `${from} - ${to}`;
if (...) fromToString.append(',')
fromToString.append('\n')

mickrobk (changes requested):

src/main/java/com/google/sps/data/OrganizationUpdater.java
Sarah-Addo: The input type used only allows a time to be inputted. If from > to then it would be a user error but that is also why we give them the option to edit their organization data

src/main/java/com/google/sps/data/OrganizationUpdater.java
mickrobk: The enum values need to be indented here. EG:
```
... class {
    enum ... {
        Monday,
        Tues
    }
}
```

src/main/java/com/google/sps/data/OrganizationUpdater.java
mickrobk: You're right that it's a user error, but in general it's important to have a good understanding of the format of the data in your datastore. (Which means validating it!) It's important to consider:
 - data you don't expect may come from malicious users! 
 - even if the use of the data is safe currently, future developers (including yourself) may make assumptions about the format of the data based on the common case and not consider that time might be "seven" or "<script>"

That all said, this seems like we can separate it from the rest of this change! What do you think about creating a github issue to track a fix for this, and addressing it in a follow up PR?

Sarah-Addo: Got it! I will create an issue and then address it

mickrobk (approved):

PR #67, opened 2020-07-20
Refactor GivrUser.java

jnnybk:
There was a logic error in GivrUser when using getCurrentLoggedInUser(). Based on our design implementation discussions, we decided that userEmail would be necessary when storing User entity in the Datastore. This means when querying to the Datastore to retrieve information (including isMaintainer status), we must query using the email (or the userId, if that exists). 

This PR removes the function getUserById and replaces it with getUserByIdOrEmail.


src/main/webapp/registration.html
```diff
@@ -4,7 +4,7 @@
 <head>
 	<meta charset="UTF-8">
 	<title>Givr</title>
-	<link rel="stylesheet" href="style.css">
+  <link rel="stylesheet" href="style.css">
```
andrewm898: you should fix the indentations here

jnnybk: I think we should fix all of the spacing (especially in the HTML files) in the Lint-Only PR!

Sarah-Addo (approved):

andrewm898 (approved):

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,18 +104,19 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
     datastore.put(entity);
   }
 
-  public static GivrUser getUserById(String userId) {
-    Entity entity = getUserFromDatastoreWithProperty("userId", userId);
+  public static GivrUser getUserByIdOrEmail(String userId, String userEmail) {
+    Entity entityRetrievedWithId = getUserFromDatastoreWithProperty("userId", userId);
+    Entity entityRetrievedWithEmail = getUserFromDatastoreWithProperty("userEmail", userEmail);
```
mickrobk: So before this change there was:
 getUserById
 getUserByEmail
 getCurrentLoggedInUser

After this change there is a:
 getUserByIdOrEmail
 getUserByEmail
 getCurrentLoggedInUser

The "before" seems like a better API to me than the after, but I understand why you want to merge these cases. What if you kept the before API, but implemented both of those functions via a new private helper: getUserByIdOrEmail

mickrobk (changes requested):

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: This makes sense! I'll add it back.

src/main/webapp/registration.html
belraquib: I think you should revert this part of the change? It looks ilke it was correct before this?

src/main/webapp/style.css
```diff
@@ -228,6 +228,11 @@
   width: 70%;
 }
 
+#add-maintainer-email-form {
```
belraquib, line 236: This change doesn't seem like it belongs in this PR. Please split these changes into different PRs for the future.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,22 +104,38 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
     datastore.put(entity);
   }
 
+  public static GivrUser getUserByIdOrEmail(String userId, String userEmail) {
```
belraquib: Do you intend to have this public or do you want this to be private?

belraquib: If you allow either userId or userEmail to be null then you can implement both getUserId and getUserByEmail below with this function.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,22 +104,38 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
...
+      isMaintainer = (boolean) entityRetrievedWithEmail.getProperty("isMaintainer");
+    }
+    
+    GivrUser user = new GivrUser(userId, isMaintainer, isLoggedIn, "" /* URL is not needed when User is logged in. */, userEmail);
+    return user;
+  }
+
   public static GivrUser getUserById(String userId) {
-    Entity entity = getUserFromDatastoreWithProperty("userId", userId);
+    Entity entityRetrievedWithId = getUserFromDatastoreWithProperty("userId", userId);
```
belraquib: Could this be simplified as


return getUserByIdOrEmail(userId, /*userEmail=*/null);

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,22 +104,38 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
...
+      isMaintainer = (boolean) entityRetrievedWithId.getProperty("isMaintainer");
+      userEmail = (String) entityRetrievedWithId.getProperty("userEmail");
     }
 
     GivrUser user = new GivrUser(userId, isMaintainer, isLoggedIn, "" /* URL is not needed when User is logged in. */, userEmail);
     return user;
   }
-
   public static GivrUser getUserByEmail(String email) {
     // TODO: Support OAuth.
```
belraquib, line 138: Could this be simplified to

return getUserByIdOrEmail(/*userId=*/null, email);

?

belraquib (changes requested):

src/main/webapp/style.css
jnnybk: I believe this change came from merge conflicts between this branch and master. Sorry!

src/main/java/com/google/sps/data/GivrUser.java
jnnybk: I do not see any uses for keeping it public, for now. I've changed this to be a private method.

jnnybk: I would then have to make getUserByIdOrEmail public since getUserByEmail is used in the OrganizationUpdater class. I decided to keep getUserByIdOrEmail as a separate (private) method and keep the other two public functions. In one of my discussions with Rob, we agreed that it would be useful to keep getUserById for a good API.

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -104,7 +104,14 @@ public static void updateUserInDatastore(String identifyingProperty, String iden
     datastore.put(entity);
   }
 
-  public static GivrUser getUserByIdOrEmail(String userId, String userEmail) {
+  private static GivrUser getUserByIdOrEmail(String userId, String userEmail) {
+    if (userId == null) {
+      userId = "";
+    } 
+    if (userEmail == null) {
+      userEmail = "";
```
belraquib: won't this match any userEmails that are empty?

Perhaps start entityRetrievedWith* as null and then set them if the corresponding input param is valid?

jnnybk: This makes sense. I made the appropriate changes to handle null userID (and userEmail).

belraquib (approved):

PR #70, opened 2020-07-21
Add 'see more results' button to paginate

andrewm898:
This adds a "See More Results" button to the bottom of the 5 results. This button disappears as soon as the last result is returned. 

image

This functionality is currently deployed, but note: Filtering will only work with zipcodes because I branched this off of master

### DIFF
--- a/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
@@ -53,7 +53,7 @@ public class ListOrganizationsServlet extends HttpServlet {
     /* All get requests will return a maximum of 5 organization entities */
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(5);
 
-    String startCursor = request.getParameter("scrs");
+    String startCursor = request.getParameter("cursor");
     if ((startCursor != null) && (!startCursor.equals("none"))) { //if the given cursor is 'none' no cursor is necessary
       fetchOptions.startCursor(Cursor.fromWebSafeString(startCursor));
     }
--- a/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
@@ -52,11 +52,22 @@ public class ListOrganizationsServlet extends HttpServlet {
 
     /* All get requests will return a maximum of 5 organization entities */
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(5);
+
+    String startCursor = request.getParameter("scrs");
+    if ((startCursor != null) && (!startCursor.equals("none"))) { //if the given cursor is 'none' no cursor is necessary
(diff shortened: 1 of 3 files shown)


src/main/webapp/style.css
```diff
@@ -82,13 +82,6 @@
   float: right;
 }
 
-#organization-list {
```
andrewm898: this class was not doing anything and removing it allows the "see more results" button to stay at the bottom of the organizations lists

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -52,11 +52,22 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
     /* All get requests will return a maximum of 5 organization entities */
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(5);
+
+    String startCursor = request.getParameter("scrs");
+    if ((startCursor != null) && (!startCursor.equals("none"))) { //if the given cursor is 'none' no cursor is necessary
+      fetchOptions.startCursor(Cursor.fromWebSafeString(startCursor));
+      System.out.println("START CURSOR HERE");
```
jnnybk: nit! remove print statement :)

jnnybk (approved):

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
andrewm898: good catch thanks

src/main/webapp/style.css
Sarah-Addo: organizations are being added to this list div. its for encapsulation and styling purposes. is there a way to refactor your code so you dont get rid of this? could you put the see more results button as the last element in the div?

src/main/webapp/search_area.js
```diff
@@ -141,11 +155,17 @@ class SearchArea {
   async getListOfOrganizations() {
     let response;
     if (this.filterParams) {
-      response = await fetch(`/list-organizations?${this.filterParams.toString()}`);
+      response = await fetch(`/list-organizations?scrs=${this.mostRecentCursor}&${this.filterParams.toString()}`);
```
belraquib: why not add this.mostRecentCursor to filterParams always?

belraquib (changes requested):

src/main/webapp/search_area.js
andrewm898: I did that, and then merged the new things in master into this PR to make sure it will work right with master's current state. I tested it with the most recent version of master and this PR is ready to merge!

src/main/webapp/search_area.js
```diff
@@ -115,6 +90,14 @@ class SearchArea {
     this.searchArea.appendChild(this.organizationSearchArea);
     this.searchArea.appendChild(this.organizationPopupArea);
   }
+  refreshOrganizationList() {
```
belraquib, line 94: add a newline between functions

belraquib (approved):

PR #73, opened 2020-07-22
add delete organization functionality

### LINKED ISSUE #71: Add delete organization button to the edit page

Sarah-Addo:
Screenshot: https://screenshot.googleplex.com/tGOQoAYLpp5

Check is failing because it uses a function from a PR in the works


src/main/webapp/organization.js
```diff
@@ -221,22 +231,6 @@ class Organization {
     orgPhoneEntry.classList.add("edit-entry");
     editForm.appendChild(orgPhoneEntry);
 
-    // label and entry area for organization zipcode
```
Sarah-Addo: Ignore, was duplicated in master before I pulled in a new branch

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,29 @@
...
 public class DeleteOrganizationServlet extends HttpServlet {
 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long id = Long.parseLong(request.getParameter("id"));
+
+    Key DistributorEntityKey = KeyFactory.createKey("Distributor", id);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.delete(DistributorEntityKey);
+
```
andrewm898, line 60: When an organization gets deleted from the datastore, should that organization's moderators be deleted from the user table if they don't moderate any other organizations?

mickrobk: Existence in the user table is unrelated to whether they're a moderator right? (So we should avoid modifying the user table based on changes to moderation status)

Sarah-Addo: Yes we decided to just call the update moderating org function for each user in the moderator list (latest commit)

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,52 @@
...
-import com.google.sps.data.GivrUser;
-import java.io.IOException;
 
+/** Servlet responsible for deleting tasks. */
 @WebServlet("/delete-organization")
 public class DeleteOrganizationServlet extends HttpServlet {
 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
belraquib, line 37: I expected a check that the user is a moderator of the organization or a maintainer here?

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,52 @@
...
+
+    ArrayList<String> moderators = (ArrayList) organizationEntity.getParameter("moderatorList");
+    datastore.delete(DistributorEntityKey);
+
+
+    // After deleting organization from datastore, go through its list of moderators and correctly update the
+    // moderating orgs they are now apart of
+    for(String userId : moderators) {
+        GivrUser user = GivrUser.getUserById(userId);
+        user.setModeratingOrgs();
```
belraquib: Do you mean updateModeratingOrgs? Probably don't need this as it's a temporary (read from the datastore, not stored) idea?

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
Sarah-Addo: Right now the organizations page only appears to moderators or maintainers and this button will only be shown on the organizations page (same with the edit button). Should there be an additional check here too though?

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
Sarah-Addo: You are right, will be removing

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,51 @@
...
 
+/** Servlet responsible for deleting tasks. */
 @WebServlet("/delete-organization")
 public class DeleteOrganizationServlet extends HttpServlet {
 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    GivrUser user = GivrUser.getCurrentLoggedInUser();
+
+    if(!user.isModeratorOfAnyOrg() || !user.isMaintainer()) {
```
belraquib: I think you want to check if the user is specifically a moderator of *this* org?

Sarah-Addo: yep, added the correct check!

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,61 @@
...
+    boolean userIsModeratorOfOrg = false;
+
+    // Check to see if the user who requested deletetion of this organization is a true moderator of the
+    // organization
+    for(Entity org : usersModeratingOrgs) {
+      long orgId = (long) org.getKey().getId();
+      if(orgId == id) {
+        userIsModeratorOfOrg = true;
+      }
+    }
```
belraquib: Can you move this into the GivrUser ?

Sarah-Addo: Got it!

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -73,6 +73,16 @@ public boolean isLoggedIn() {
     return this.isLoggedIn;
   }
 
+  public boolean isModeratorOfOrganization(long organizationId) {
+    for(Entity entity : this.moderatingOrgs) {
```
belraquib: I think you should call this.getModeratingOrgs() here, instead of this.moderatingOrgs, otherwise the field may not be populated yet. Can you add a comment on the field too (or make it private?)

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,51 @@
...
-import java.io.IOException;
 
+/** Servlet responsible for deleting tasks. */
 @WebServlet("/delete-organization")
 public class DeleteOrganizationServlet extends HttpServlet {
 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    long id = Long.parseLong(request.getParameter("id"));
```
belraquib: nit: call this organizationId ?

src/main/java/com/google/sps/data/GivrUser.java
Sarah-Addo: make getModeratingOrgs private?

belraquib: I meant this.moderatingOrgs (not getModeratingOrgs()). The idea is that way all callers are forced to call getModeratingOrgs; which will ensure they don't make the same mistake. Up to you.

Sarah-Addo: okay I added a comment above moderating orgs to let users know the getModeratingOrgs should be called for access

src/main/java/com/google/sps/servlets/DeleteOrganizationServlet.java
```diff
@@ -14,14 +14,51 @@
...
 public class DeleteOrganizationServlet extends HttpServlet {
 
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    long organizationId = Long.parseLong(request.getParameter("id"));
+    
+    GivrUser user = GivrUser.getCurrentLoggedInUser();
+
+    if(!user.isModeratorOfOrganization(organizationId) || !user.isMaintainer()) {
```
belraquib: I think you mean

!user.isModeratrOfOrg() && !isMaintainer

a test will catch this

belraquib (approved):

PR #74, opened 2020-07-22
Fix semantics of resource listing

andrewm898:
Changed the listing on the resources to be from a JSON stringify to a comma delimited array, and changes category to categories, since it should be plural

old: https://screenshot.googleplex.com/niW4ZdmgvKD

new: 
image


jnnybk (approved):
I like this change!

mickrobk (approved):
Nice!

PR #80, opened 2020-07-24
Create Event class

jnnybk:
This PR sets up the Event Java class based on slide 21 in Data Structuring presentation: https://docs.google.com/presentation/d/1lGDiZT6umMahr-Jmqd8VLkIsO9dTW2PJkN3eCDV68PY/edit?usp=sharing.

The reason for choosing Map<Date, EmbeddedEntity> to represent the event's date(s) and hours was:
- an Event can span multiple days and have different hours for individual dates.

(EDIT) 

This PR creates Address class and Contact class. 

(EDIT 2)

Removed Address class and Contact class as separate classes based on pod's discussion with the interest of time + ease of use + simplicity of code


Sarah-Addo (approved):

andrewm898 (approved):

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,41 @@
...
+package com.google.sps.data;
+
+import java.util.Map;
+import java.util.Date;
+import java.util.ArrayList;
+import com.google.appengine.api.datastore.EmbeddedEntity;
+
+public class Event {
+
+  private long id;
```
belraquib, line 26: comments for each field, please. I'm confused why some ids are strings and some are ints and some are embedded entities.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,43 @@
...
+package com.google.sps.data;
+
+import java.util.Map;
+import java.util.Date;
+import java.util.ArrayList;
+import com.google.appengine.api.datastore.EmbeddedEntity;
+
+public class Event {
+
+  private long id; // Represents ID of Event; Type is long because Datastore's Key's getId() return type is long.
```
belraquib: nit: put the comment above the field, instead of next to.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,43 @@
...
+  private String title; // Represents the title of the Event.
+  private ArrayList<Long> ownerOrgIds; // Holds IDs (type long) of Organizations that own the Event.
+  private ArrayList<EmbeddedEntity> partnerIdsOrNames; // Holds EmbeddedEntity that represents an Organization's ID OR name. A partnering organization could have missing ID.
+  private String description; // Represents description of the Event. e.g. "In front of church ABC"
+  private Map<Date, EmbeddedEntity> dateAndHours; // Holds dates of Event and their corresponding time(s). 
+  private String contactEmail; // Represents a single contact email for this Event.
```
belraquib: I wonder if there's a more generic Contact structure that can be used here and in the organizations as well. I can imagine contacts may also need names / terms of address?

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,43 @@
...
+  private String contactPhone; // Represents a single contact phone number for this Event.
+
+  // Represents location information for the Event.
+  private String address;
```
belraquib: Is there an opportunity to use a more generic Address structure for this and the Organizations?

belraquib (changes requested):

src/main/java/com/google/sps/data/Event.java
belraquib: Consider: https://cloud.google.com/appengine/docs/standard/python/datastore/typesandpropertyclasses#PostalAddress

belraquib: java 8 version:
https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PostalAddress

https://cloud.google.com/appengine/docs/standard/java/datastore/entity-property-reference

src/main/java/com/google/sps/data/Event.java
jnnybk: I will discuss this with the pod and update based on the discussion!

src/main/java/com/google/sps/data/Event.java
jnnybk: Our pod's concern is that we would need to store the zipcode separately or somehow parse this (if all addresses were in the same format). We've considered using PostalAddress but it seems more appropriate to store these attributes separately.

src/main/java/com/google/sps/data/Event.java
jnnybk: After the pod discussion, we've agreed to keep it as it is because we don't feel the need to store address with the contact information.

I think that creating a contact class would be an option, too.

src/main/java/com/google/sps/data/Event.java
belraquib: That is valid - I hadn't realized it was unparsed! That all being said, consider creating your own Address class, so that there isn't a chance of the representation here skewing from the organization (or maybe you _do_ want that)

src/main/java/com/google/sps/data/Event.java
belraquib: When I said terms of address I meant more the "I want to be called bob".

jnnybk: Sorry! That makes a lot more sense. 

Would it be better to create a contact class having attributes for 
1) first name 
2) last name
3) phone number
4) email address,

or to have separate fields within the Event class holding this information?

belraquib: The former (and you may want to do this for Organization contacts, if and when you have that as well)

jnnybk: Okay, sounds good! Thank you!

I will make an issue to refactor Organization class since this pull request focuses on Event class! :)

src/main/java/com/google/sps/data/Event.java
belraquib: Can you do what you did for contact here?

Address {
   street_address
   city
   state
   zipcode
}

?

jnnybk: Yes :)

src/main/java/com/google/sps/data/Address.java
```diff
@@ -0,0 +1,31 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+public class Address {
+
+  private String address;
```
belraquib: Add comments to each field that may be ambiguous or may need specifically formatted data (ie, only numbers). What do you expect in each? Eg. Is address the street + apt number or would that be in a separate field?

src/main/java/com/google/sps/data/Contact.java
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+public class Contact {
+  // Represents contact email.
```
belraquib: Instead of repeating the field name, consider giving an example. 

Any email address, e.g.  help@theawesomeplace.com

src/main/java/com/google/sps/data/Contact.java
```diff
@@ -0,0 +1,36 @@
...
+  private String email;
+  // Represents contact phone number.
```
belraquib: Instead of/in addition to calling out that this is a phone number, tell the reader what to expect here:

A 10 digit US number: e.g. 111 111 1111 or the text equivalent, 800 HELP ALL1

?

src/main/java/com/google/sps/data/Contact.java
```diff
@@ -0,0 +1,36 @@
...
+  private String phone;
+  // Represents contact first name.
+  private String firstName;
+  // Represents contact last name.
+  private String lastName;
```
belraquib: are both of these required? Can they be left empty? The comment is where you should specify your preconditions

belraquib (changes requested):
Getting there! I'm being very nitpicky, but the documentation is important, as it lets everyone get on the same page

src/main/java/com/google/sps/data/Address.java
jnnybk: I added these comments and an example!

jnnybk:
UPDATE: Removed Address class and Contact class as separate classes based on pod's discussion with the interest of time.

belraquib (approved):

PR #81, opened 2020-07-24
Remove generation of login URL

jnnybk:
Only generate login url if the user is not logged in.

### DIFF
--- a/src/main/java/com/google/sps/data/GivrUser.java
+++ b/src/main/java/com/google/sps/data/GivrUser.java
@@ -144,9 +144,9 @@ public class GivrUser {
     
     if (isUserLoggedIn) {
       return getUserById(userService.getCurrentUser().getUserId());
-    } else {
-      url = userService.createLoginURL("/");
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -140,9 +140,10 @@ public static GivrUser getUserByEmail(String email) {
   public static GivrUser getCurrentLoggedInUser() {
     UserService userService = UserServiceFactory.getUserService();
     boolean isUserLoggedIn = userService.isUserLoggedIn();
-    String url = userService.createLoginURL("/");
+    String url = "";
     
     if (isUserLoggedIn) {
+      url = userService.createLoginURL("/");
```
belraquib: erm I had misrecalled before; you only want to create the url if this is !isUserLoggedIn, right?

jnnybk: yes. would this mean removing the generation completely?

belraquib (approved):

Issue #85, opened 2020-07-27
Events backend


jnnybk:
This issue tracks our last main feature's back-end work, Events.

1. Event class - PR #80 
2. Event Updater class - Skeleton set in PR #86, Implementation in PR #96, Refactoring OrganizationUpdater in PR #101 
3. Add Event Servlet - PR #86

PR #86, opened 2020-07-27
Add AddEventServlet and EventUpdater class skeleton

jnnybk:
This PR completes Issue #85 no.3 and sets up EventUpdater's skeleton class.

AddEventServlet accomplishes the following:
- Check if user logged in and is a Moderator
- Creates a new Event entity
- Updates the Event entity with correct information by using EventUpdater
- Sends redirect to /index.html

NOTE: Should be merged AFTER PR #80 is merged.


jnnybk:
NOTE: This PR fails the build check because it is missing Event.java and EventUpdater.java (PR #80).

Sarah-Addo (approved):

andrewm898 (approved):

src/main/java/com/google/sps/servlets/AddEventServlet.java
```diff
@@ -0,0 +1,68 @@
...
+      throw new IllegalArgumentException("ERROR: Unable to register event if user is not logged in.");
+    }
+
+    // Only Moderators can add Events.
+    if (user.isModeratorOfAnyOrg()) {
+      // TODO: Can Maintainers add Events?
+      throw new IllegalArgumentException("ERROR: User does not have the permission to perform addition of Events.");
+    }
+
+    Entity newEventEntity = new Entity("Event");
```
belraquib, line 48: For now, can you come up with what you expect the API to be from the point of view of the servlet? What are the fields that are going to be sent in the form. Minimally i expect the owning organization id. You can then check your auth against that.

PR #87, opened 2020-07-27
create events page and additional things we will need for it

### LINKED ISSUE #94: Create event front end page

Sarah-Addo:
Create main events page. 
Screenshot: https://screenshot.googleplex.com/1Gmx1Q5VY3L


src/main/webapp/user.js
```diff
@@ -28,9 +28,9 @@ class User {
 
     if (loginData.isLoggedIn) {
       this.isMaintainer = loginData.isMaintainer;
+      this.isModerator =  this.userIsModerator();
```
jnnybk: PR #59 was merged, so this should be updated! :)

src/main/webapp/search_area.js
```diff
@@ -16,32 +16,52 @@ document.addEventListener("DOMContentLoaded", async function() {
...
+      const eventsSearchArea = new SearchArea(document.getElementById('events'), isMaintainer, forOrganizationsPage, forEventsPage);  }
 });
 
 class SearchArea {
-  constructor(searchAreaElement, isMaintainer, forOrganizationsPage) {
-    this.searchArea = searchAreaElement;
-    this.organizationSearchArea = document.createElement("div");
+  constructor(searchAreaElement, isMaintainer, forOrganizationsPage, forEventsPage) {
+    this.searchAreaContainer = searchAreaElement;
+    this.searchArea = document.createElement("div");
```
andrewm898: could you update this on line 47 of filter_tag_area.js?

Sarah-Addo: yup its updated

src/main/webapp/search_area.js
```diff
@@ -16,32 +16,52 @@ document.addEventListener("DOMContentLoaded", async function() {
...
     this.isMaintainer = isMaintainer;
+    // TODO(): Add isModerator check after moderator PR is pulled in
+    this.isModerator = true;
     this.forOrganizationsPage = forOrganizationsPage;
+    this.forEventsPage = forEventsPage;
     this.filterParams = new URLSearchParams();
     this.organizationObjectsList = [];
+    this.eventObjectsList = [];
```
andrewm898: Since this code never operates in a situation where organizationObjectsList AND eventObjectsList need to have values at the same time, what if we generalized organizationObjectsList/eventObjectsList to be something like "objectsList", and then the events and organizations are both filled in there? This way, we would only need to change lines 145 and 174 to use different classes for the 2 situations, and different servlets, respectively.

We would also be able to do the same thing with organizationListArea / eventListArea: we could reference the list area by a single name in the code at places like line 132, 162, etc. but just assign that list area in the constructor to either be an event list area or org list area.

If we can generalize the repeated objects into using the same ones for events / orgs, it will be very easy to connect this to the backend

src/main/webapp/index.html
```diff
@@ -18,8 +18,7 @@
 			<div id="site-title">Givr</div>
 			<div id="nav-links">
 				<a href="index.html">Get Help</a>
-				<a href="registration.html">Register Organization</a>
```
belraquib: Why is the registration being removed?

src/main/webapp/search_area.js
Sarah-Addo: okay!

src/main/webapp/index.html
Sarah-Addo: It should have been removed when we implemented the rebuild nav bar method. Registration option is only shown after the user has logged in

src/main/webapp/search_area.js
```diff
@@ -16,32 +16,52 @@ document.addEventListener("DOMContentLoaded", async function() {
   // This checks that requester User has valid credentials to edit/delete/view ALL organizations. (by checking userID)
   let isMaintainer = false;
   let forOrganizationsPage = false;
-
+  let forEventsPage = false;
   if (document.getElementById('search-area')) {
```
belraquib: This continues to be very odd for me. I believe the plan was to create a new SearchArea in js for each individual page, so you don't have the breakouts? This is otherwise leaking element ids?

src/main/webapp/search_area.js
```diff
@@ -16,32 +16,52 @@ document.addEventListener("DOMContentLoaded", async function() {
   // This checks that requester User has valid credentials to edit/delete/view ALL organizations. (by checking userID)
   let isMaintainer = false;
   let forOrganizationsPage = false;
```
belraquib: Is the expectation that only one of organizations page / events page is set?

src/main/webapp/search_area.js
```diff
@@ -16,32 +16,52 @@ document.addEventListener("DOMContentLoaded", async function() {
...
+    this.forEventsPage = forEventsPage;
     this.filterParams = new URLSearchParams();
     this.organizationObjectsList = [];
+    this.eventObjectsList = [];
     
     /* The cursor is the "cursor" filter param, and the keyword "none" is used to start at the beginning of the query */
     this.filterParams.set("cursor", "none");
     this.lastResultFound = false;
 
+    if(forEventsPage && (this.isModerator || this.isMaintainer)) {
```
belraquib: Why do you want to show this only in the events page?

src/main/webapp/search_area.js
```diff
@@ -120,53 +150,92 @@ class SearchArea {
...
     this.filterParams.set("cursor", newCursor);
     /* If < 5 results are returned, the end of the given query has been reached */
-    this.lastResultFound = (this.organizationObjectsList.length < 5);
+    this.lastResultFound = (this.objectsList.length < 5);
     if (this.lastResultFound) {
       this.loadMoreButton.classList.add("hide-load-button");
     }
   }
 
+  renderListOfEvents() {
```
andrewm898: We can condense the render / get functions into one function for each & just decide which class / fetch link to use inside of those functions instead of line 128

andrewm898: we can take care of this when we connect this to the backend though fine to leave as is for this PR

Sarah-Addo: I think for reuseablity purposes that they should be separate in the case that we want to render organizations differently than events, thoughts?

src/main/webapp/events_search_area.js
```diff
@@ -0,0 +1,92 @@
...
+    this.myEventsButton.textContent = "My Events";
+    this.myEventsButton.addEventListener('click', () => this.addMyEventsFilter());
+    this.myEventsAndAddButtons.appendChild(this.myEventsButton);
+    this.addEventButton = document.createElement("a");
+    this.addEventButton.textContent = "Register an Event";
+    this.addEventButton.setAttribute("href", "register_event.html");
+    this.myEventsAndAddButtons.appendChild(this.addEventButton);
+    this.searchArea.appendChild(this.myEventsAndAddButtons);
+
+    this.searchAreaObject = new SearchArea(this.searchArea, (objectsList, listArea) => { return this.renderListOfEvents(objectsList, listArea)}, (filterParams, objectsList, lastResultFound, loadMoreButton) => { return this.getListOfEvents(filterParams, objectsList, lastResultFound, loadMoreButton)});
```
belraquib: can you clean up the lint here a bit?

```
new SearchArea(this.searchArea,
                           (..) => { 
                               ...;
                           },
                           (..) => {...});
```

src/main/webapp/index.html
```diff
@@ -19,8 +19,7 @@
 			<div id="site-title">Givr</div>
 			<div id="nav-links">
 				<a href="index.html">Get Help</a>
-				<a href="register_organization.html">Register Organization</a>
-				<!--this will only appear when logged in after login is setup -->
+				<a href="events.html">Events</a>
```
belraquib, line 22: Why is register org being removed from the index page?

belraquib (changes requested):
Thanks ! This is much better scoped

src/main/webapp/index.html
belraquib: NVM - saw your previous response.

src/main/webapp/organization_search_area.js
```diff
@@ -32,7 +32,9 @@ class OrganizationSearchArea {
     this.searchArea = searchAreaElement;
     this.isMaintainer = isMaintainer;
     this.forOrganizationsPage = forOrganizationsPage;
-    this.searchAreaObject = new SearchArea(this.searchArea, (objectsList, listArea) => { return this.renderListOfOrganizations(objectsList, listArea)}, (filterParams, objectsList, lastResultFound, loadMoreButton) => { return this.getListOfOrganizations(filterParams, objectsList, lastResultFound, loadMoreButton)});
+    this.searchAreaObject = new SearchArea(this.searchArea,
+      (objectsList, listArea) => { return this.renderListOfOrganizations(objectsList, listArea) },
+      (filterParams, objectsList, lastResultFound, loadMoreButton) => { return this.getListOfOrganizations(filterParams, objectsList, lastResultFound, loadMoreButton) });
```
belraquib: put the return on a new line, here and above

belraquib (approved):

PR #89, opened 2020-07-27
create event registration form

### LINKED ISSUE #93: Create event registration page

Sarah-Addo:
Screenshot: https://screenshot.googleplex.com/1m5ufnNCr80

### DIFF
--- a/src/main/webapp/open_hours.js
+++ b/src/main/webapp/open_hours.js
@@ -15,26 +15,27 @@
 document.addEventListener("DOMContentLoaded", async function() {
   // Looks for the hours-option-area in dom and will append the option to enter times 
   // for when the organization is open on every day of the week
+  const optionArea = document.getElementById("hours-option-area");
   if (document.getElementById("registration-form-area")) {
-    const optionArea = document.getElementById("hours-option-area");
-    const mondayTimeOption = new TimeOption("Monday", true, null, optionArea);
-    const tuesdayTimeOption = new TimeOption("Tuesday", true, null, optionArea);
-    const wednesdayTimeOption = new TimeOption("Wednesday", true, null, optionArea);
-    const thursdayTimeOption = new TimeOption("Thursday", true, null, optionArea);
-    const fridayTimeOption = new TimeOption("Friday", true, null, optionArea);
-    const saturdayTimeOption = new TimeOption("Saturday", true, null, optionArea);
-    const sundayTimeOption = new TimeOption("Sunday", true, null, optionArea);
+    const mondayTimeOption = new TimeOption("Monday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const tuesdayTimeOption = new TimeOption("Tuesday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const wednesdayTimeOption = new TimeOption("Wednesday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const thursdayTimeOption = new TimeOption("Thursday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const fridayTimeOption = new TimeOption("Friday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const saturdayTimeOption = new TimeOption("Saturday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const sundayTimeOption = new TimeOption("Sunday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
   }
 });
 
 class TimeOption {
   // organization.hoursOpen[index(1-7 for day of week)].propertyMap.
   // fromToPairs.value[index(how many set of hours for that day )].propertyMap.from/to
-  constructor(day, forRegistration, organizationDay, optionArea) {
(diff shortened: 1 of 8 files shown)


src/main/webapp/open_hours.js
```diff
@@ -45,36 +50,51 @@ class TimeOption {
...
-      this.dayOpenInput.setAttribute("checked", "checked");
-      this.timeInputArea.classList.add("show-time-area");
-    } else if (!this.organizationDay.propertyMap.isOpen) {
-      this.dayClosedInput.setAttribute("checked", "checked");
-      this.timeInputArea.classList.add("hide-time-area");
-    } else {
-      this.dayOpenInput.setAttribute("checked", "checked");
-      this.timeInputArea.classList.add("show-time-area");
+
+    // Only show open closed options for organization modification
```
jnnybk: I'm not sure if having this for specific pages and not for other pages is the best way to solve reusing the TimeOption class, if we decide to go with this option.

For example, I was thinking about having the Event registration page having:

(MM/DD/YYYY)
Date                    Time(s)
07/20/2020         3:15 - 4:15
                           6:15 - 7:15
07/21/2020         5:00 - 7:00

Sarah-Addo: Right now it im designing it mvp style which would mean that the event would only have one day as an option so it is easier to implement. But they can have multiple times for that day

andrewm898 (approved):

src/main/webapp/open_hours.js
```diff
@@ -45,36 +50,51 @@ class TimeOption {
...
+    if (!document.getElementById("event-registration-form-area")) {
```
belraquib: Generally special casing if a certain id is available is a bad code smell.

You have several options here; if you want to do so: 
   move this code into a :   renderIntoElement(element) {}  function, that will allow you decide to create the elements in the calling code. You can also extract a new class that does this instead that encapsulates the data as well as creates the elements.

belraquib (changes requested):

src/main/webapp/register_event.html
```diff
@@ -0,0 +1,47 @@
...
+				<input class = "registration-entry" type = "text" name = "event-state"/>
+				<label>Event Zip Code: </label>
+				<input class = "registration-entry" type = "text" name = "event-zip-code"/>
+				<label>Contact Email: </label>
+				<input class = "registration-entry" type = "text" name = "event-email"/>
+				<label>Contact Phone Number: </label>
+				<input class = "registration-entry" type = "text" pattern = "[0-9]{10}" name = "event-phone-num"/>
+				<label>Event Details: </label>
+				<textarea class = "registration-entry" name ="event-details" maxlength = "180"></textarea>
+				<label>Event Partners: </label>
```
jnnybk, line 39: I think we should also have an Event Owners field, based on our Data Structuring powerpoint discussion!

src/main/webapp/register_event.html
```diff
@@ -0,0 +1,47 @@
...
+				<input class = "registration-entry" type = "text" name = "event-street-address"/>
+				<label>Event City:</label>
+				<input class = "registration-entry" type = "text" name = "event-city"/>
+				<label>Event State:</label>
+				<input class = "registration-entry" type = "text" name = "event-state"/>
+				<label>Event Zip Code: </label>
+				<input class = "registration-entry" type = "text" name = "event-zip-code"/>
+				<label>Contact Email: </label>
+				<input class = "registration-entry" type = "text" name = "event-email"/>
+				<label>Contact Phone Number: </label>
```
jnnybk, line 37: We should have a Contact Name

src/main/webapp/open_hours.js
belraquib: Let's continue to think about this:

I believe we talked about the options of : extract creating these open/close options in a new function (and call it elsewhere!) or pass in a constructor argument that tells this object whether or not to create these elements.

src/main/webapp/open_hours.js
```diff
@@ -45,36 +46,51 @@ class TimeOption {
...
-      this.dayOpenInput.setAttribute("checked", "checked");
-      this.timeInputArea.classList.add("show-time-area");
-    } else if (!this.organizationDay.propertyMap.isOpen) {
-      this.dayClosedInput.setAttribute("checked", "checked");
-      this.timeInputArea.classList.add("hide-time-area");
-    } else {
-      this.dayOpenInput.setAttribute("checked", "checked");
-      this.timeInputArea.classList.add("show-time-area");
+
+    // Only show open closed options for organization modification
```
belraquib: update the comment here

src/main/webapp/open_hours.js
```diff
@@ -17,24 +17,25 @@ document.addEventListener("DOMContentLoaded", async function() {
...
+    const saturdayTimeOption = new TimeOption("Saturday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
+    const sundayTimeOption = new TimeOption("Sunday", /*forRegistration*/ true, null, optionArea, /*showOpenClosedOptions*/ true);
   }
 });
 
 class TimeOption {
   // organization.hoursOpen[index(1-7 for day of week)].propertyMap.
   // fromToPairs.value[index(how many set of hours for that day )].propertyMap.from/to
-  constructor(day, forRegistration, organizationDay, optionArea) {
+  constructor(day, forRegistration, organizationDay, optionArea, showOpenClosedOptions) {
```
belraquib, line 33: Leave a small comment for each parameter.

belraquib (approved):

PR #90, opened 2020-07-27
Separate list servlets

andrewm898:
This PR separates ListOrganizationsServlet into ListHelper, and adds ListEventsServlet to handle listing events

Notes: 
-ListOrganizationServlet and ListEventsServlet both have a 'constantMap' hashmap. These are used so the same filtering logic can be applied over their shared, but differently named fields (think orgZipCode vs eventZipCode, etc.)

-moderatingOrgs was made public so the test case can assign it to a hardcoded value without needing to add a setter to the class

-This implementation is using a form of the Event.java file that does not implement the address and contact class- I left those out of this for now to avoid making this PR too long.

- Test cases for the event servlet specifically are in the PR after this one.


src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -85,83 +95,4 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     response.getWriter().println(gson.toJson(requestedOrganizations));
     response.addHeader("Cursor", encodedEndCursor);
   }
-
```
andrewm898, line 88: This function was separated into 2 and moved into ListHelper

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,57 @@
...
+    this.partnerIdsOrNames = (ArrayList) entity.getProperty("eventPartnerOrgIdsOrNames");
+    this.description = (String) entity.getProperty("eventDescription");
+    this.contactEmail = (String) entity.getProperty("eventEmail");
+    this.address = (String) entity.getProperty("eventStreetAddress");
+    this.city = (String) entity.getProperty("eventCity");
+    this.state = (String) entity.getProperty("eventState");
+    this.zipcode = (String) entity.getProperty("eventZipCode");
+    this.dateAndHours = (Map) entity.getProperty("dateAndHours");
+    this.contactPhone = (String) entity.getProperty("eventPhoneNum");
+    this.creationTimeStampMillis = (long) entity.getProperty("creationTimeStampMillis");
```
Sarah-Addo: The timestamps are not apart of the java file since this is what is getting returned to the frontend for usage and these are only used for the backend

src/main/java/com/google/sps/data/GivrUser.java
```diff
@@ -47,7 +47,7 @@
   private String url;
   private String email;
   // In method setModeratingOrgs(), ArrayList will be set with all Organizations that this GivrUser moderates.
-  private ArrayList<Entity> moderatingOrgs = new ArrayList<Entity>();
+  public ArrayList<Entity> moderatingOrgs = new ArrayList<Entity>();
```
Sarah-Addo: why is this public? shouldn't it be private and have a get method?

andrewm898: In the test cases I'm assigning a hardcoded array directly to this variable since "setModeratingOrgs" is already implemented in this class (and I explicitly don't want to use it in the test cases, the array needs to be hardcoded there). I figured it would be confusing to create another "setModeratingOrgs" function, so I made it so that the variable can just be assigned directly

edit: changed it back to be private, didn't actually need it to be public

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,55 @@
...
+package com.google.sps.data;
+
+import java.util.Map;
+import java.util.Date;
+import java.util.ArrayList;
+import com.google.appengine.api.datastore.EmbeddedEntity;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EmbeddedEntity;
+
+public class Event {
```
jnnybk: sorry #80 is taking a while, but it should be in soon!

Sarah-Addo (approved):

src/main/java/com/google/sps/servlets/ListEventsServlet.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.IOException;
+
+@WebServlet("/list-events")
+public class ListEventsServlet extends HttpServlet {
+
+  /* Events & orgs have different labels for fields (ex: orgCity vs eventCity). This map is used to reference
+   * the event-specific fields easily from ListHelper */
+  public static HashMap<String, String> constantMap = new HashMap<String, String>();
+  static {
+    constantMap.put("name", "eventTitle");
```
jnnybk: update: eventTitle->eventName

I've updated Event class to reflect changes in the front-end!

src/main/java/com/google/sps/servlets/ListOrganizationsServlet.java
```diff
@@ -48,6 +49,15 @@
 @WebServlet("/list-organizations")
 public class ListOrganizationsServlet extends HttpServlet {
 
+  /* Events & orgs have different labels for fields (ex: orgCity vs eventCity). This map is used to reference
```
belraquib: why? can we change this?

src/main/java/com/google/sps/servlets/ListEventsServlet.java
```diff
@@ -0,0 +1,98 @@
...
+import com.google.sps.data.ListHelper;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.sps.data.GivrUser;
+import java.io.IOException;
+
+@WebServlet("/list-events")
+public class ListEventsServlet extends HttpServlet {
+
+  /* Events & orgs have different labels for fields (ex: orgCity vs eventCity). This map is used to reference
```
belraquib: if all the fields were named the same, this would be obviated. Can we do that?

src/main/java/com/google/sps/servlets/ListEventsServlet.java
```diff
@@ -0,0 +1,98 @@
...
+  public static HashMap<String, String> constantMap = new HashMap<String, String>();
+  static {
+    constantMap.put("name", "eventName");
+    constantMap.put("streetAddress", "eventStreetAddress");
+    constantMap.put("zipcode", "eventZipCode");
+  }
+
+  /*
+   * This get request returns a list of events depending on its query parameters. 
+   * If no parameters are included, it will return a default list of organizations
```
belraquib: update comments. here and elsewhere :)

src/main/java/com/google/sps/data/ListHelper.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.IOException;
+
+public class ListHelper {
+
+  public static Query getQuery(String entityKind, HttpServletRequest request, GivrUser currentUser) {
+    
+    /* allows generic word to reference specific field for given entity kind ("name" -> "orgName", etc.) */
+    HashMap<String, String> datastoreConstantMap;
+    if (entityKind.equals("Distributor")) {
+      datastoreConstantMap = new HashMap<String, String>(ListOrganizationsServlet.constantMap);
```
belraquib: This introduces a cycle. Let's remove it :)

belraquib (changes requested):

src/main/java/com/google/sps/data/ListEventsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.sps.data.GivrUser;
+import com.google.sps.servlets.ListEventsServlet;
+import com.google.sps.data.ListHelper;
+import java.io.IOException;
+
+public class ListEventsHelper extends ListHelper {
+
+  public static Query getQuery(String entityKind, HttpServletRequest request, GivrUser currentUser) {
+    
+    HashMap<String, String> datastoreConstantMap = new HashMap<String, String>(ListEventsServlet.constantMap);
```
belraquib: can you move the constant map into this class? That will change the circular dependency of:

ListEventsServlet -> ListEventsHelper -> ListEventsServlet.constantMap

Looks like it's only used here anyway.

src/main/java/com/google/sps/data/ListEventsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /* First the filter params in the request are parsed into a map that applies each filter to the query */
+    HashMap<String, ArrayList<String>> filterParamMap = parseFilterParams(entityKind, request, currentUser, datastoreConstantMap);
```
belraquib: you no longer need entity kind here, right?

src/main/java/com/google/sps/data/ListEventsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+    ArrayList<Filter> filterCollection = new ArrayList<Filter>();
+    filterCollection = handleUserFiltering(entityKind, currentUser, displayForUser);
+
+    return getQueryFromFilters(entityKind, filterParamMap, filterCollection);
+  }
+
+  /* handleUserFiltering handles filtering related to a user's role and permissions, and whether that user has requested to only
+   * see organizations or events they belong to */
+  public static ArrayList<Filter> handleUserFiltering(String entityKind, GivrUser currentUser, boolean displayForUser) {
```
belraquib: also don't need it here

src/main/java/com/google/sps/data/ListEventsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.appengine.api.users.UserServiceFactory;
+import javax.servlet.http.HttpServletRequest;
+import com.google.sps.data.GivrUser;
+import com.google.sps.servlets.ListEventsServlet;
+import com.google.sps.data.ListHelper;
+import java.io.IOException;
+
+public class ListEventsHelper extends ListHelper {
+
+  public static Query getQuery(String entityKind, HttpServletRequest request, GivrUser currentUser) {
```
belraquib: You can simplify this now, since you know that this is always the "Events" kind you don't need it as a parameter.

src/main/java/com/google/sps/data/ListEventsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    HashMap<String, String> datastoreConstantMap = new HashMap<String, String>(ListEventsServlet.constantMap);
+
+    /* First the filter params in the request are parsed into a map that applies each filter to the query */
+    HashMap<String, ArrayList<String>> filterParamMap = parseFilterParams(entityKind, request, currentUser, datastoreConstantMap);
+
+    /* displayForUser is set to true when the user wants whatever entity they are querying
+     * (Distributors, Events), to only return the entities they belong to / are involved with.*/
+    boolean displayForUser = coerceParameterToBoolean(request, "displayForUser");
```
belraquib: Looks like this will always result in an empty list if displayForUser is not true. Is that your intent? Can you return early in that case?

belraquib (changes requested):

src/main/java/com/google/sps/data/ListHelper.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Query.CompositeFilter;
+import com.google.appengine.api.datastore.Query.CompositeFilterOperator;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import javax.servlet.http.HttpServletRequest;
+import com.google.sps.data.GivrUser;
+import java.io.IOException;
+
+public abstract class ListHelper {
```
belraquib, line 44: Argh... github keeps losing my comments!

While this works, it's not quite what I was thinking. The class here only has static functions; you don't need to mark the class as abstract in that case (the above specializations can just call the static functions).

For posterity, here's what I was thinking:

```

public abstract class ListHelper {
  protected String entityKind;
  protected HttpServletRequest request;
  protected GivrUser currentUser;

  ListHelper(String entityKind, HttpServletRequest request, GivrUser currentUser) {
    // store all args...
  }

  public Query getQuery() {
    return this.getQueryFromFilters(
        this.handleUserFiltering(coerceParameterToBoolean(this.request, "displayForUser")));
  }

  // Impemented by the subclass.
  // Gets the constant map that translates from known key -> datastore property for the entity.
  abstract HashMap<String, String> GetDatastoreConstantMap();
  // Returns the filter for this.entity against this.user
  abstract ArrayList<Filter> handleUserFiltering(boolean displayForUser);

  // Common functions used for query.
  private HashMap<String, ArrayList<String>> parseFilterParams() {
    // old code with the following changes:
    // datastore map = this.GetDatastoreConstantMap();
    // you can get the request from: this.request
  }

  /* This function constructs a query based on the entity kind, request parameters and any previously added or new filters */
  private Query getQueryFromFilters()
      throws IllegalArgumentException {
    // same as what you have, except you don't pass in args.
    // filterParamMap = this.parseFilterParams();
    // everything else is in instance vars.
  }

  public boolean coerceParameterToBoolean(HttpServletRequest String key) {...}
}

public class ListEventsHelper extends ListHelper {
  HashMap<String, String> GetDatastoreConstantMap() {
    // return a new instance of the constant map here instead of in the servlet.
  }

  public ArrayList<Filter> handleUserFiltering(boolean displayForUser) {
    // basically your old functions, but use the instance vars.
  }
}

public class ListOrganizationsHelper extends ListHelper {
  HashMap<String, String> GetDatastoreConstantMap() {
    // return a new instance of the constant map here instead of in the servlet.
  }

  public ArrayList<Filter> handleUserFiltering(boolean displayForUser) {
    // basically your old functions, but use the instance vars.
  }
}


```

belraquib (changes requested):

src/main/java/com/google/sps/data/ListEventsHelper.java
andrewm898: yes- my intent here is to only add this filter to the query if the user requested it, and if not just leave the filterCollection empty & continue adding to it in the other filtering function.

src/main/java/com/google/sps/data/ListHelper.java
```diff
@@ -0,0 +1,181 @@
...
+  protected HttpServletRequest request;
+  protected GivrUser currentUser;
+
+  ListHelper(String entityKind, HttpServletRequest request, GivrUser currentUser) {
+    this.entityKind = entityKind;
+    this.request = request;
+    this.currentUser = currentUser;
+  }
+
+  public Query getQuery() {
```
belraquib, line 56: Leave a comment on all public functions as to what you expect it to do :)

src/main/java/com/google/sps/data/ListHelper.java
```diff
@@ -86,8 +107,75 @@ public static Query getQueryFromFilters(String entityKind, HashMap<String, Array
...
 
-  public static boolean coerceParameterToBoolean(HttpServletRequest request, String key) {
+  public boolean coerceParameterToBoolean(HttpServletRequest request, String key) {
     String requestParameter = request.getParameter(key);
     return (requestParameter != null) && (requestParameter.equals("true"));
   }
 }
+
+
+// public abstract class ListHelper {
```
belraquib: did you intend to leave this in?

belraquib (changes requested):
LGTM just nits left

belraquib (approved):

PR #91, opened 2020-07-27
Add event test cases

andrewm898:
This adds test cases for the event servlets in the previous PR

Note: build will fail because that code is not in this PR

### DIFF
--- a/src/test/java/com/google/sps/ListEventServletTest.java
+++ b/src/test/java/com/google/sps/ListEventServletTest.java
@@ -187,4 +187,7 @@ public final class ListEventServletTest {
     FetchOptions fetchOptions = FetchOptions.Builder.withLimit(10);
     Assert.assertArrayEquals(expectedList.toArray(), datastore.prepare(receivedQuery).asList(fetchOptions).toArray());
   }
+
+  // TODO: add a test for when the user is a Maintainer, when the user is a moderator of an org with
+  // no events, and when the user is not a moderator
 }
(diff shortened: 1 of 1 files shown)


src/test/java/com/google/sps/ListEventServletTest.java
```diff
@@ -0,0 +1,155 @@
...
+ 
+  private DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+ 
+  /* The masterEntityList is a list of entities that are populated into the local datastore, and also exist 
+   * in an arraylist. For each test, the datastore applies the Query received from the servlet, and then
+   * compares that query to a hardcoded list of expected entities given that tests parameters / user fields */
+  private ArrayList<Entity> masterEntityList = new ArrayList<>();
+ 
+  @Before
+  public void setUp() {
```
Sarah-Addo, line 59: how come we are only testing if the zipcode works and not sorting by other properties?

andrewm898: Right now this test is just making sure that the event listing for a specific user works. I agree it can use a couple more tests- will add those now!

Sarah-Addo (approved):

src/test/java/com/google/sps/ListEventServletTest.java
```diff
@@ -0,0 +1,188 @@
...
+    helper.tearDown();
+  }
+
+  @Test
+  public void testEventQueryForUserOrgs() {
+    /* Tests to make sure that client can successfully request only user's events */
+    GivrUser mockUser = mock(GivrUser.class);
+    
+    ArrayList<Entity> moderatingOrgs = new ArrayList<Entity>();
+    /* This user moderates the orgs w/ ID's 1 and 3 */
```
belraquib, line 142: note that this is entity list items 0, 1, and 5

belraquib (approved):
This looks really good! I will approve now, but can you also add a test for when the user is a Maintainer, when the user is a moderator of an org with no events, and when the user is not a moderator?

PR #95, opened 2020-07-28
refactor search area and add organization_search_area class

Sarah-Addo:


andrewm898 (approved):

jnnybk (approved):

src/main/webapp/organization_search_area.js
```diff
@@ -0,0 +1,81 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+document.addEventListener("DOMContentLoaded", async function() {
+  // This checks that requester User has valid credentials to edit/delete/view ALL organizations. (by checking userID)
+  let isMaintainer = false;
+  let forOrganizationsPage = false;
+
+  if (document.getElementById('search-area')) {
```
belraquib, line 20: This is looking great! Question: why do we have a if/else here?

src/main/webapp/search_area.js
```diff
@@ -67,106 +51,49 @@ class SearchArea {
...
+    this.loadMoreButton = document.createElement("div");
+    this.loadMoreButton.setAttribute("class", "load-more-button");
+    this.loadMoreButton.textContent = "See More Results";
+    this.loadMoreButton.addEventListener('click', () => this.handleObjects());
+
+    this.listArea = document.createElement("div");
+    this.popupArea = document.createElement("div");
+    console.log(this.forOrganizations);
+    if (this.forOrganizations) {
+      this.listArea.setAttribute("id", "organization-list");
```
belraquib: can you call this 

search-result-list and then not have the if/else ? I think that means you can then remove this.forOrganizations

src/main/webapp/search_area.js
```diff
@@ -67,106 +51,49 @@ class SearchArea {
...
+  refreshObjectsList() {
+    this.filterParams.set("cursor", "none");
+    this.lastResultFound = false;
+    this.loadMoreButton.classList.remove("hide-load-button");
+    this.objectsList = [];
+    this.listArea.innerHTML = "";
+    this.handleObjects();
   }
 
-  async setUrlParamValue(urlParamKey, urlParamValue) {
```
belraquib, line 152: I think you want to keep this?

src/main/webapp/search_area.js
```diff
@@ -67,106 +51,49 @@ class SearchArea {
...
+    this.loadMoreButton.setAttribute("class", "load-more-button");
+    this.loadMoreButton.textContent = "See More Results";
+    this.loadMoreButton.addEventListener('click', () => this.handleObjects());
+
+    this.listArea = document.createElement("div");
+    this.popupArea = document.createElement("div");
+    console.log(this.forOrganizations);
+    if (this.forOrganizations) {
+      this.listArea.setAttribute("id", "organization-list");
+      this.popupArea.setAttribute("id", "organization-popup-area");
```
belraquib: same here:

result-popup-area :)

src/main/webapp/organization_search_area.js
Sarah-Addo: It looks like it is to decide if we are creating the search area for the organizations tab. I think it is to solve the issue of if we are showing the edit/delete buttons or not

belraquib: IMO we can go one further and a different JS for organizations.html (organizations_page.js) and index.html (landing_page.js)

That gets rid of these if statements, breaking the cycle of the the organization search area know about the page it is being rendered on. I'm only proposing moving/changing lines 15-28 here.

Feel free to do that in another PR though!

belraquib (approved):
This looks much nicer! Thanks.

PR #96, opened 2020-07-28
Implement EventUpdater Java class

jnnybk:
This PR implements EventUpdater used to create and edit Events.

(EDIT) Must be merged AFTER #101


andrewm898 (approved):

Sarah-Addo (approved):

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +49,159 @@ public Entity getEntity() {
...
+    this.entity.setProperty(propertyKey, formValue);
   }
+
+  // Sets values not passed in through the request form such as creation time of Event and last edited time.
+  private void setNonFormProperties(boolean forRegistration, EmbeddedEntity historyUpdate) {
+    long milliSecondsSinceEpoch = (long) historyUpdate.getProperty("changeTimeStampMillis");
+    ArrayList<EmbeddedEntity> changeHistory = new ArrayList<EmbeddedEntity>();
+
+    if (forRegistration) {
+      this.entity.setProperty("eventCreationTimeStampMillis", milliSecondsSinceEpoch);
```
andrewm898: can you remove the 'event' from the beginning of the timestamp properties?

jnnybk: yes!

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +49,159 @@ public Entity getEntity() {
     return this.entity;
   }
 
+  // TODO: When editing event (after event has been registered), we must check if the user sending the request has the right credentials - add GivrUser.isModeratorOfOrgWithId(long organizationId)
```
belraquib: why is this a TODO? Any reason you can't do that now?

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +49,159 @@ public Entity getEntity() {
...
+      String formKey = entry.getKey();
+      String formValue = "";
+
+      if (optionalProperties.contains(propertyKey)) {
+        formValue = request.getParameter(formKey) == null ? "" : request.getParameter(formKey);
+      } else {
+        try {
+          formValue = getParameterOrThrow(request, formKey);
+        } catch (IllegalArgumentException err) {
+          logger.log(Level.SEVERE, "Form value for: " + propertyKey + " cannot be left blank.");
```
belraquib, line 112: should you re-raise the exception here? Alternatively, don't handle the exception in this function and let the servlet handle it.

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +49,159 @@ public Entity getEntity() {
...
+
+    Filter queryFilter = new FilterPredicate(Entity.KEY_RESERVED_PROPERTY, FilterOperator.EQUAL, ownerOrgId);
+    Query query = new Query("Distributor").setFilter(queryFilter);
+    PreparedQuery preparedQuery = datastore.prepare(query);
+
+    Entity entity = null;
+    try {
+      entity = preparedQuery.asSingleEntity();
+    } catch(PreparedQuery.TooManyResultsException exception) {
+      logger.log(Level.SEVERE, "Multiple Distributor entities found with ID: " + Long.toString(ownerOrgId) + ".");
```
belraquib: Anywhere there is an exception, you should bubble it up to the servlet. Otherwise this will look like the update succeeded.

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +49,159 @@ public Entity getEntity() {
...
+    }
+
+    String ownerOrgName = (String) entity.getProperty("orgName");
+    this.entity.setProperty("eventOwnerOrgName", ownerOrgName);
+  }
+
+  private String getParameterOrThrow(HttpServletRequest request, String formKey) {
```
belraquib: I remember seeing this function and the one below elsewhere. Can you make a comment request handling library?

src/main/java/com/google/sps/data/EventUpdater.java
belraquib: Doesn't look like the change happened?

belraquib (changes requested):

src/main/java/com/google/sps/data/EventUpdater.java
jnnybk: It's complete; we're figuring out connecting front+backend so I'll have it pushed soon!

jnnybk:
Build WILL FAIL - Must be merged after #101

src/main/java/com/google/sps/data/Event.java
```diff
@@ -27,16 +27,18 @@
   private long id;
   // Represents the name of the Event.
   private String name;
-  // Holds ID (type long) of the Organization that owns the Event.
+  // Holds ID (type String) of the Organization that owns the Event.
```
belraquib: is it a string or a long?

src/main/java/com/google/sps/data/Event.java
```diff
@@ -27,16 +27,18 @@
...
+  private String ownerOrgName;
   /* Holds ArrayList of Strings that represents an Organization's name. 
    * No need to store IDs of partnering organization, as partners will not have the privilege of editing Events.
    */
   private ArrayList<String> partnerNames;
   // Represents details of the Event. e.g. "In front of church ABC"
   private String details;
-  // Holds Date of Event and their corresponding time(s).
-  private HashMap<Date, EmbeddedEntity> dateAndHours;
+  // Holds Date of Event and their corresponding time(s). EmbeddedEntity holds a Date and an ArrayList of EmbeddedEntities that represents the time range(s) for that specific Date. This was chosen over a Map because Datastore does not support Maps.
```
belraquib: Is the purpose of this class to be convenient for:
    converting to gson?
    use in the servlets?
    modifying the datastore?

I would say that on class likely can't do all 3. If this object was for converting to gson, then the data and hours might actually be better used as a map. Let's call that out at the top.

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +53,163 @@ public Entity getEntity() {
...
+
+    if (organizationEntity == null) {
+      throw new IllegalArgumentException("There is no Organization with ID: " + orgId);
+    }
+
+    return organizationEntity;
+  }
+
+  private boolean doesUserHasCredentialsToUpdateEvent(GivrUser user, long orgId) {
+    Entity entity = getOrgEntityWithId(orgId);
```
belraquib: I thought there was a 
GiverUser.isModeratorOfOrg(long orgId) {} function. If there isn't can you make one?

belraquib (changes requested):

src/main/java/com/google/sps/data/Event.java
jnnybk: long! Sorry, fixing this right now.

src/main/java/com/google/sps/data/Event.java
jnnybk: I believe this class is used for all three things listed. For the backend (use in servlet and modifying in datastore), it seems to make most sense to use ArrayList of embedded entities, especially because Datastore cannot store HashMaps. For the frontend, it does make more sense to use HashMaps, but we seem to not run into any problems when converting ArrayList of embedded entities using Gson. 

I'll add a comment!

src/main/java/com/google/sps/data/EventUpdater.java
jnnybk: Added! I moved getOrgEntityWithId into Organization class since it doesn't deal with updating Organization. Is this okay?

jnnybk:
NOTE: Must be merged after #101, still!

src/main/java/com/google/sps/data/Event.java
```diff
@@ -29,14 +29,16 @@
...
+  private String ownerOrgName;
   /* Holds ArrayList of Strings that represents an Organization's name. 
    * No need to store IDs of partnering organization, as partners will not have the privilege of editing Events.
    */
   private ArrayList<String> partnerNames;
   // Represents details of the Event. e.g. "In front of church ABC"
   private String details;
-  // Holds Date of Event and their corresponding time(s).
-  private HashMap<Date, EmbeddedEntity> dateAndHours;
+  // Holds Date of Event and their corresponding time(s). EmbeddedEntity holds a Date and an ArrayList of EmbeddedEntities that represents the time range(s) for that specific Date. This was chosen over a Map because Datastore does not support Maps and will be mainly used in servlets + modifying data in Datastore. For converting to Json using Gson in ListEventsServlet, it would be more appropriate to store this as a HashMap.
```
belraquib: I think you can elide the last sentence.

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +54,137 @@ public Entity getEntity() {
...
+     *
+     * The following properties are optional:
+     * - eventPartnerNames
+     * - eventDetails
+     */
+    Set<String> optionalProperties = new HashSet<String>();
+    optionalProperties.add("eventPartnerNames");
+    optionalProperties.add("eventDetails");
+
+    for (Map.Entry<String, String> entry: formProperties.entrySet()) {
```
belraquib, line 97: I almost want to extract this into the request handler helper as well (this for loop i mean). You would have to pass in a lambda for the final set form value call. I'm imagining:
```
// pseudocode
ParserHelper.SetPropertiesFromForm(
      request,    
      formProperties,
      function(String propertyKey, String formValue) {
           setEventProperty(propertyKey, formValue);
      });
```

belraquib (approved):

src/main/java/com/google/sps/data/EventUpdater.java
jnnybk: Would this function also handle the other logic in the for loop, inside of the setPropertiesFromForm function?

jnnybk: I'm going to add this into the Issue list, because there are a couple of PRs waiting for this!

PR #101, opened 2020-07-30
Create ParserHelper and RequestHandler

jnnybk:
This PR pulls out three functions:
1) getParameterOrThrow
2) getParameterValuesOrThrow
3) createFromToPairs
from OrganizationUpdater and adds 1) + 2) to ParserHelper.java / 3) to RequestHandler.java.

Purpose: These three functions are used in EventUpdater.

PR #106, opened 2020-07-31
Field update

andrewm898:
Updates fields according to this key: https://docs.google.com/document/d/1cs6RxZugRQvWojcaJCEfwG4TtRnFYP9mHEo46I7D5xM/edit

NOTE: this PR contains changes from (at the time of opening this PR) the other open PR's #96 and #101, since those files had fields that needed to be updated. This will be merged after those ones are


Sarah-Addo (approved):

jnnybk (approved):
looks good, it seems that there are irrelevant files included in this PR!

src/main/java/com/google/sps/data/EventUpdater.java
```diff
@@ -36,7 +53,159 @@ public Entity getEntity() {
     return this.entity;
   }
 
-  public void updateEvent(HttpServletRequest request, GivrUser user, boolean forRegistration, EmbeddedEntity historyUpdate) throws IllegalArgumentException {
+  // Called when creating and updating Event from AddEventServlet and EditEventServlet.
+  public void updateEvent(HttpServletRequest request, GivrUser user, boolean forRegistration, EmbeddedEntity historyUpdate) throws IllegalArgumentException {
```
belraquib: ... this diff looks wrong? rebase? IIRC, i think this is from Jenny's change?

andrewm898: I included this file because there were fields in here that needed to be updated as well, Jenny's change with it hasn't been merged yet though

PR #109, opened 2020-07-31
Fix retrieval of isModerator status in OrganizationUpdater

jnnybk:
This PR accomplishes retrieving the user's moderator status based on GivrUser's isModeratorOfOrgWithId(long orgId). Before this change, the isModerator status of the user was being set true even if the user was not a moderator of the specific org that is being updated.


Sarah-Addo (approved):

belraquib (commented):
Is there a test you can update with this change?

jnnybk:
> Is there a test you can update with this change?

I'm currently writing tests for this! :)

belraquib (approved):
Happy to approve this now and have the test in another PR

PR #111, opened 2020-08-03
Finish edit event functionality

Sarah-Addo:
Accomplishes:
- Getting correct moderator and maintainer status for events
- Completes full edit functionality for events

Screenshot: https://screenshot.googleplex.com/PQ7feWxUH2v

### DIFF
--- a/src/main/java/com/google/sps/servlets/EditEventServlet.java
+++ b/src/main/java/com/google/sps/servlets/EditEventServlet.java
@@ -38,14 +38,9 @@ public class EditEventServlet extends HttpServlet {
     GivrUser user = GivrUser.getCurrentLoggedInUser();
 
     if (!user.isLoggedIn()) {
-      throw new IllegalArgumentException("ERROR: Unable to register event if user is not logged in.");
+      throw new IllegalArgumentException("ERROR: Unable to edit event if user is not logged in.");
     }
 
-    // Only Moderators can add Events.
-    if (!user.isModeratorOfAnyOrg() && !user.isMaintainer()) {
-      throw new IllegalArgumentException("ERROR: User does not have the permission to perform addition of Events.");
-    }
-    
     //Get the proper event entity for updating
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     long eventId = Long.parseLong(request.getParameter("id"));
--- a/src/main/java/com/google/sps/servlets/EditEventServlet.java
+++ b/src/main/java/com/google/sps/servlets/EditEventServlet.java
@@ -60,7 +60,8 @@ public class EditEventServlet extends HttpServlet {
     HistoryManager history = new HistoryManager();
     EmbeddedEntity historyUpdate = history.recordHistory("Event was edited.", millisecondsSinceEpoch);
 
-    // Update rest of Event properties from request form.
+    // Update rest of Event properties from request form. Will not update event if user that
+    // requested is not a moderator for the event or a maintainer
     try {
       eventUpdater.updateEvent(request, user, /* forRegistration */ false, historyUpdate);
     } catch (IllegalArgumentException err) {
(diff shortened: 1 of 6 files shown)


jnnybk (approved):
lgtm : )

src/main/java/com/google/sps/servlets/EditEventServlet.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    GivrUser user = GivrUser.getCurrentLoggedInUser();
+
+    if (!user.isLoggedIn()) {
+      throw new IllegalArgumentException("ERROR: Unable to register event if user is not logged in.");
+    }
+
+    // Only Moderators can add Events.
+    if (!user.isModeratorOfAnyOrg() && !user.isMaintainer()) {
```
belraquib: should this check that the user is an editor of _this_ org?

belraquib (changes requested):

andrewm898 (approved):

src/main/java/com/google/sps/servlets/EditEventServlet.java
Sarah-Addo: the correct check was actually already in line 67 of event updater so i just removed this check from the servlet

src/main/java/com/google/sps/servlets/EditEventServlet.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    EventUpdater eventUpdater = new EventUpdater(eventEntity);
+    long millisecondsSinceEpoch = Instant.now().toEpochMilli();
+    HistoryManager history = new HistoryManager();
+    EmbeddedEntity historyUpdate = history.recordHistory("Event was edited.", millisecondsSinceEpoch);
+
+    // Update rest of Event properties from request form.
+    try {
+      eventUpdater.updateEvent(request, user, /* forRegistration */ false, historyUpdate);
```
belraquib, line 66: Can you verify and add a comment that this checks to see if the user is allowed to update the event?

src/main/webapp/event.js
```diff
@@ -364,20 +376,56 @@ class Event {
...
   }
+
+  parseDate(dateAndHours) {
+    // Will change format from MM DD, YYYY 00:00AM to MM DD, YYYY
+    const date = dateAndHours.propertyMap.date.slice(0, -12);
+    return date;
+  }
+
+  parseDateForValue(dateAndHours) {
+    // Will translate a date in format MM DD, YYYY to YYYY-MM-DD for purpose of prefilling input date value
```
belraquib, line 403: Oof...  I would normally recommend using https://momentjs.com/  for this (and if you want to try, go for it!)

belraquib, line 403: leave a comment that this is the format from the java representation of a Date

src/main/webapp/open_hours.js
```diff
@@ -100,7 +100,13 @@ class TimeOption {
 
     // for edit page get the number of pairs of time input options to preset the values
     let numPairs = 0;
-    if (!forRegistration && this.organizationDay.propertyMap.isOpen) {
+    if (!forRegistration && !this.showOpenClosedOptions) {
+      numPairs = this.organizationDay.propertyMap.fromToPairs.value.length;
+
+      // set initial from to pair, if organization is open there is at least one pair
+      this.dayFromInput.setAttribute("value", this.organizationDay.propertyMap.fromToPairs.value[0].propertyMap.from);
```
belraquib: can you store propertyMap into a temp variable for readability?

belraquib (changes requested):
mostly nits

src/main/webapp/event.js
Sarah-Addo: I think I am going to keep it as is for now for the sake of time but I will definitely use it in the future!

belraquib (approved):

PR #112, opened 2020-08-04
filtering bug fix

andrewm898:
Resolves bug related to having multiple active filters for singleton properties. Before, the user could add multiple address / name filters, however this flow does not make sense since the backend applies the filters with the AND operator, so multiple filters of one property will always return no results. The only way to apply multiple filters on the same property is with composite OR operators, however as was previously found, this limits the total number of filters that can be applied.

Now, the "Example Org B" filter will be replaced by the Example Org A filter once it is entered:
image


jnnybk (approved):

Sarah-Addo (approved):

src/main/webapp/search_area.js
```diff
@@ -95,18 +95,19 @@ class SearchArea {
...
-      this.filterParams.set(urlParamKey, urlParamValue);	
+    
+    /* Only resource categories can have multiple active filters */
+    if (urlParamKey === "resourceCategories") {	
+      this.filterParams.append(urlParamKey, urlParamValue);
     } else {	
-      this.filterParams.append(urlParamKey, urlParamValue);	
+      if (this.filterParams.get(urlParamKey)) {
+        /* If there is an active param for this key, remove its tag so both aren't displayed */
+        this.filterTagArea.removeFilterTag(urlParamKey, this.filterParams.get(urlParamKey), document.getElementById(urlParamKey));
```
belraquib: can you explicitly pass /* refreshObjects= */ false a the last param?

src/main/webapp/filter_tag.js
```diff
@@ -36,7 +33,7 @@ class FilterTag {
     this.filterTagClose = document.createElement("div");
     this.filterTagClose.textContent = 'X';
     this.filterTagClose.setAttribute("class", "filter-tag-close");
-    this.filterTagClose.addEventListener('click', () => this.parentFilterTagArea.removeFilterTag(urlParamKey, urlParamValue, this));
+    this.filterTagClose.addEventListener('click', () => this.parentFilterTagArea.removeFilterTag(urlParamKey, urlParamValue, this, true));
```
belraquib: /* refreshObjects= */ true

belraquib (changes requested):

belraquib (approved):

PR #118, opened 2020-08-06
More tests

andrewm898:
Adds tests for ListEventHelper for when the user is a Maintainer, when the user is a moderator of an org with no events, and when the user is not a moderator. Also fixes bug on ListEventHelper to remove error that happens when moderator id list is an empty collection

### DIFF
--- a/src/test/java/com/google/sps/ListEventServletTest.java
+++ b/src/test/java/com/google/sps/ListEventServletTest.java
@@ -188,9 +188,6 @@ public final class ListEventServletTest {
     Assert.assertArrayEquals(expectedList.toArray(), datastore.prepare(receivedQuery).asList(fetchOptions).toArray());
   }
 
-  // TODO: add a test for when the user is a Maintainer, when the user is a moderator of an org with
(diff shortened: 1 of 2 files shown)


jnnybk (approved):

src/test/java/com/google/sps/ListEventServletTest.java
```diff
@@ -188,6 +188,75 @@ public void testEventQueryForFilteredUserOrgs() {
...
+ 
+    ListEventsHelper listEventsHelper = new ListEventsHelper("Event", mockRequest, mockUser);
+    Query receivedQuery = listEventsHelper.getQuery();
+
+    FetchOptions fetchOptions = FetchOptions.Builder.withLimit(10);
+    Assert.assertArrayEquals(expectedList.toArray(), datastore.prepare(receivedQuery).asList(fetchOptions).toArray());
+  }
+
+  @Test
+  public void testEventQueryForOrgWithoutEvents() {
```
belraquib, line 218: can you add a test for an org with events?

belraquib (approved):

